PR #72, opened 2017-11-21
Fix missing version headers

sam-token:


lahosken:
This fix seems legit. If the version headers have been missing, any idea why bamboo hasn't been screaming at us?

sam-token:
@lahosken the Gateway validator only throws if the header is set and not valid. If it is absent it just [logs a warning](https://github.com/tokenio/gateway/blob/master/src/main/java/io/token/gateway/rpc/service/VersionValidator.java#L55).

lahosken:
D'oh! Good catch, sir.

PR #73, opened 2017-11-21
recovery samples. some api additions to get samples working

lahosken:
sample code showing: setting up recovery for default recovery; recovering via default; setting up non-default; recovering via non-default

tweaked sdk behavior for verifyAlias; update SDK's cache of aliases if verifyAlias is succeeding.

creating an authorization needs member's lastHash; added an API to fetch member info and put its lastHash into an authorization.

### DIFF
--- a/lib/src/main/java/io/token/MemberAsync.java
+++ b/lib/src/main/java/io/token/MemberAsync.java
@@ -295,7 +295,11 @@ public final class MemberAsync {
      * @return a completable
      */
     public Completable verifyAlias(String verificationId, String code) {
-        return client.verifyAlias(verificationId, code);
+        Completable done = client.verifyAlias(verificationId, code);
+        done.blockingAwait();
+        // We probably have at least one new alias; refresh alias cache.
+        this.aliases = new ArrayList<>(client.getAliases().blockingSingle());
+        return done;
     }
 
     /**
--- a/lib/src/main/java/io/token/MemberAsync.java
+++ b/lib/src/main/java/io/token/MemberAsync.java
@@ -297,11 +297,7 @@ public final class MemberAsync {
      * @return a completable
      */
     public Completable verifyAlias(String verificationId, String code) {
-        Completable done = client.verifyAlias(verificationId, code);
-        done.blockingAwait();
-        // We probably have at least one new alias; refresh alias cache.
-        this.aliases = new ArrayList<>(client.getAliases().blockingSingle());
-        return done;
+        return client.verifyAlias(verificationId, code);
     }
 
     /**
--- a/samples/src/main/java/io/token/sample/MemberRecoverySample.java
+++ b/samples/src/main/java/io/token/sample/MemberRecoverySample.java
@@ -1,0 +1,140 @@
+package io.token.sample;
+
+import static io.token.proto.common.security.SecurityProtos.Key.Level.PRIVILEGED;
+
+import io.token.Member;
+import io.token.TokenIO;
+import io.token.proto.common.alias.AliasProtos.Alias;
+import io.token.proto.common.member.MemberProtos.MemberRecoveryOperation;
+import io.token.proto.common.member.MemberProtos.MemberRecoveryOperation.Authorization;
+import io.token.proto.common.member.MemberProtos.RecoveryRule;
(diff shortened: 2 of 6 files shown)


lahosken:
Reviewers, some things to be extra-careful about:

For the non-default recovery sample, I didn't use our default agent at all. I'm not 100% sure that's allowed. When I peek at the code, I ?think? it sends a code to the user's email. So maybe I should always user our default agent as one of the agents? ¯\_(ツ)_/¯

I ?think? I had to add an SDK api so I could get the member's lastHash to put into the recovery Authority. But if that makes you say "Hey no... there's a way to get that, you didn't need to add a method" that would be good to know.

I called that method `createRecoveryAuthorization` and it creates an authorization. Maybe there's a better API to provide? Like maybe there should be TokenIO.fetchLastHash(memberId) instead? Or something? Or maybe there's a better name for this thing?

samples/src/main/java/io/token/sample/MemberRecoverySample.java
```diff
@@ -0,0 +1,140 @@
...
+    public Member recoverWithComplexRule(
+            TokenIO tokenIO,
+            Alias alias) {
+        // complexRecovery begin snippet to include in docs
+        String memberId = tokenIO.getMemberId(alias);
+
+        CryptoEngine cryptoEngine = new TokenCryptoEngine(memberId, new InMemoryKeyStore());
+        Key newKey = cryptoEngine.generateKey(PRIVILEGED);
+
+        String verificationId = tokenIO.beginRecovery(alias);
```
grace09thompson: Something I hadn't considered before: thinking on the case where a member _doesn't_ want to use Token as a recovery agent, I can see it going two ways: first, verifying an alias like you have here and simply skipping the 'completeRecovery' part that returns a signed authorization package and going straight to 'verifyAlias', or second, bypassing the alias verification here, and having to re-verify their aliases after they have recovered. However, I am not sure how the second scenario would actually work, since our app doesn't currently allow someone to have multiple aliases, and there is no indicator for verified/unverified. Guessing eventually you would be able to select a marked-unverified alias and initiate the verification process, but not sure. I'm thinking something like that will have to be required though, since if the user had 3 aliases, our recovery process will only re-verify one of them so they will undoubtedly need a way to re-verify the others. Safest way is definitely to re-verify the alias while recovering, since the idea of having only unverified aliases is a bit frightening.

samples/src/main/java/io/token/sample/MemberRecoverySample.java
```diff
@@ -0,0 +1,140 @@
...
+            Alias agentAlias) {
+        // Someday in the future, this user might ask the recovery agent
+        // "Please tell Token that I am the member with ID m:12345678 ."
+        // While we're setting up this new member, we need to tell the
+        // recovery agent the new member ID so the agent can "remember" later.
+        tellRecoveryAgentMemberId(newMember.memberId());
+
+        String agentId = tokenIO.getMemberId(agentAlias);
+        RecoveryRule recoveryRule = RecoveryRule.newBuilder()
+                .setPrimaryAgent(agentId)
```
yuhanniu: I think the rule should always have a primary agent. And the primary agent could or could not be Token, both ways should work.

lahosken:
"Regarding the last hash, I wonder how the app retrieves the last hash when performing member updates?"

D'oh, I thought it cached this when logging in/createMembering/updating. But now I see there's a lastHash() method to fetch the hash. I swear I looked for it before. Will fix.

lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -295,7 +295,11 @@ public Completable useDefaultRecoveryRule() {
      * @return a completable
      */
     public Completable verifyAlias(String verificationId, String code) {
-        return client.verifyAlias(verificationId, code);
+        Completable done = client.verifyAlias(verificationId, code);
```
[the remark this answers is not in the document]
lahosken: Sounds good to me.

lahosken:
I retract my previous "D'oh"--that lastHash method is on Member, which the SDK will make if you create a member or log in, but not if all you have is a memberId.

PR #74, opened 2017-11-22
Members don't cache aliases, fetch by RPC each time

lahosken:


lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -146,17 +140,26 @@ public String apply(MemberProtos.Member member) throws Exception {
      * @return first alias owned by the user
      */
     @Nullable
-    public Alias firstAlias() {
-        return aliases.isEmpty() ? null : aliases.get(0);
+    public Observable<Alias> firstAlias() {
```
lahosken: not sure if that Nullable should still be there

yuhanniu: I don't think it should. It would be returning Observable of null instead of a null pointer. Although I am thinking, would it be better if we can return Observable.empty() here?

PR #79, opened 2017-12-05
login name is confusing, so change it

lahosken:
A developer had some reasonable questions about `TokenIO.login`, paraphrased: "...but there's no API to log out. What happens when a member is logged in? Is a session established? OK to login the same member several times?" Colin points out that we could change the "login" name to something less confusing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,7 @@
 If you're upgrading to a new version of the SDK,
 see this list for Java APIs that have changed:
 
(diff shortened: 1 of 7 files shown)


mariano54:
We should probably change objc and sdk-js as well for consistency. (also includes web apps and apps. ouch :P)

lahosken:
Agreed. I figured I'd let the name-bikeshedding play out on this issue first.

lahosken:
Nice. method name is now `getMember`

PR #83, opened 2017-12-13
Bl-119: Implementation of payloads for getbalance/gettransaction requests

VKCR:
The new GetBalance/GetTransaction(s) calls return a status along with the actual results to indicate whether the call was a success or if more signatures were needed. The current SDK's methods for these calls only return the result, not the status. I added Get(...)Response which return the whole proto response message so that the status can be used by developers, but I kept the old methods too.


mariano54:
@VKCR Is this still valid?

VKCR:
Yes

mariano54:
You need to update the build.gradle file with new version, otherwise the new artifact will not get published.

VKCR:
All the old method calls didn't return the status. I kept them for backwards compatibility and added the annotation deprecated. As for why there were two calls for getBalance (current and available), I don't know, it was already like that.

lahosken:
> "why there were two calls for getBalance (current and available)"

Story time:

In the beginning there was no current/available, there was just "balance."

Then we realized we needed current-vs-available in some context. So we tweaked our services. And we tweaked the sdk getBalance method to ignore one of the balances. (I forget which we kept and which we ignored. ANYHOW...)

Eventually, we realized sdk users might want to know current-vs-available, so added another sdk method.

colin-token:
Isn't it kind of weird for us to be returning the response object from the RPC directly? What if we change the gateway interface for the RPC call? Then the SDK users will need to update as well? It's best to maintain a level of abstraction between the communication protocol and the objects the SDK interacts with.

Also in other cases we throw Java exceptions when there is a problem with the status right? Any reason we don't do that anymore and expose the GRPC status to the SDK users directly here?

VKCR:
I was mimicking the way endorseToken works. Granted endorseToken doesn't return the actual grpc response but rather a TokenOperationResult message. I suppose we could define "getBalanceResult" and "getTransactionResult" but honestly they'd be identical to the current getBalanceResponse/... messages we have now. 

I like the idea of throwing exceptions, I think it would work well. This way we could even keep the current signatures of the GetBalance/GetTransaction methods in Member and Account. I'll open a new ticket.

PR #85, opened 2017-12-18
Delete keys

mariano54:


lib/src/main/java/io/token/security/TokenCryptoEngine.java
```diff
@@ -56,6 +56,19 @@ public TokenCryptoEngine(String memberId, KeyStore keyStore) {
                 .cryptoFor(CRYPTO_TYPE);
     }
 
+    /**
+     * Creates an instance, and specify whether to delete member's keys, or not.
+     *
+     * @param memberId member ID
+     * @param keyStore key store
```
lahosken: `@param delete purge member's existing keys`

mariano54:
Yes, I added the deleteKeys method to InMemoryKeyStore.

lahosken:
D'oh, yep. So you did

lib/src/main/java/io/token/security/InMemoryKeyStore.java
```diff
@@ -66,4 +69,12 @@ public SecretKey getById(String memberId, String keyId) {
     public List<SecretKey> listKeys(String memberId) {
         return new ArrayList<>(keys.row(memberId).values());
     }
+
+    @Override
+    public void deleteKeys(String memberId) {
+        Set<String> memberKeys = new HashSet<>(keys.row(memberId).keySet());
```
akalini: Is there no keys.clear() method?

lib/src/main/java/io/token/security/TokenCryptoEngine.java
```diff
@@ -56,6 +56,20 @@ public TokenCryptoEngine(String memberId, KeyStore keyStore) {
...
+     * @param delete purge member's existing keys
+     */
+    public TokenCryptoEngine(String memberId, KeyStore keyStore, boolean delete) {
```
akalini: Why would one pass in delete flag instead of calling keyStore.deleteKeys(...)?

lib/src/test/java/io/token/security/KeyStoreTest.java
```diff
@@ -39,6 +39,12 @@ public void putThenGet() throws Exception {
         store.put("steve", phone);
         assertThat(store.getById("steve", "phone")).isEqualTo(phone);
         assertThat(store.getById("steve", "laptop")).isEqualTo(laptop);
+
+        assertThat(store.listKeys("steve").size()).isEqualTo(2);
+        store.deleteKeys("steve");
+
```
akalini: remove empty line.

lib/src/main/java/io/token/security/InMemoryKeyStore.java
mariano54: Yes but here we are removing only the keys for one member

lib/src/main/java/io/token/security/TokenCryptoEngine.java
mariano54: The developer is using TokenCryptoEngine but doesn't necesarily want to deal with the keystore object directly right? It's private field

lib/src/test/java/io/token/security/KeyStoreTest.java
```diff
@@ -39,6 +39,8 @@ public void putThenGet() throws Exception {
         store.put("steve", phone);
         assertThat(store.getById("steve", "phone")).isEqualTo(phone);
         assertThat(store.getById("steve", "laptop")).isEqualTo(laptop);
+
```
akalini: Can we test the new methods?

mariano54: Yes

lib/src/main/java/io/token/security/InMemoryKeyStore.java
```diff
@@ -28,7 +28,10 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
```
tienn: looks unused

PR #89, opened 2018-01-09
Add new SDK methods

mariano54:


lib/src/main/java/io/token/TokenIO.java
```diff
@@ -271,6 +271,18 @@ public NotifyStatus notifyPaymentRequest(TokenPayload tokenPayload) {
                 .blockingSingle();
     }
 
+    /**
+     * Sends a notification to request a payment.
```
lahosken: copy-pasted? maybe: Sends a notification that an access token has expired.

lib/src/main/java/io/token/Member.java
```diff
@@ -837,6 +841,26 @@ public Money getCurrentBalance(String accountId) {
...
+        return async.triggerStepUpNotification(stepUp).blockingSingle();
+    }
+
+    /**
+     * Triggers a step up notification for an information request.
+     *
+     * @param stepUp notification payload
+     * @return status of the notification
+     */
+    public NotifyStatus triggerStepUpNotification(RequestStepUp stepUp) {
```
lahosken: I don't understand how this is supposed to work. I think that's because I don't understand the proto interface. Maybe a sample or a test would help?

lahosken: on another PR, VKCR sent me the explanation, yay: Article 10 page 16: http://ec.europa.eu/finance/docs/level-2-measures/psd2-rts-2017-7782_en.pdf

PR #92, opened 2018-01-11
Browser API

gianluca-token:
1. Provide Browser API interface (Browser.java) and Android implementation, which will open a WebView in a new Activity to display the pages.
2. linkAccounts method which takes in browser, directs the browser through bank pages, and completes entire linking process, returning linked accounts.

Usage example with Android SDK:
``` java
 BrowserFactory browserFactory = new TokenBrowserFactory(MainActivity.this);
 BankAuthorization auth = sdk.linkAccounts(bankInfo, browserFactory);
```

### DIFF
--- a/lib/src/main/java/io/token/util/Util.java
+++ b/lib/src/main/java/io/token/util/Util.java
@@ -42,6 +42,7 @@ import io.token.proto.common.member.MemberProtos.MemberOperationMetadata.AddAlia
 import io.token.proto.common.security.SecurityProtos.Key;
 
 import java.util.concurrent.ExecutionException;
+import javax.annotation.Nullable;
 
 
 /**
@@ -77,6 +78,7 @@ public abstract class Util {
 
     /**
      * Get alias with normalized value. E.g. "Captain@gmail.com" to "captain@gmail.com".
+     *
      * @param rawAlias { EMAIL, "Captain@gmail.com" }
      * @return alias with possibly-different value field
      */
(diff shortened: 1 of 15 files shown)


android/src/main/java/io/token/browser/WebViewBrowser.java
```diff
@@ -0,0 +1,28 @@
+package io.token.browser;
+
+import io.reactivex.Observable;
+import io.reactivex.subjects.Subject;
+
+import java.net.URL;
+
+/**
+ * Android WebView browser implementation.
```
tienn: is it not yet implemented or is it stub?

lib/src/main/java/io/token/util/Util.java
```diff
@@ -193,4 +198,25 @@ public void run() {
...
+
+    /**
+     * Loads the body of the resource associated with the specified url.
+     *
+     * @param url the url to fetch
+     * @return the body of the resource
+     * @throws IOException if there was a problem loading the url
+     */
+    public static String fetchUrl(URL url) throws IOException {
+        try(InputStream is = url.openConnection().getInputStream()) {
```
tienn: try is not needed here?

android/src/main/java/io/token/browser/WebViewBrowser.java
gianluca-token: It was a stub, became TokenBrowserFactory in a later commit

PR #94, opened 2018-01-12
Fingerprint hybrid

mariano54:
Lowers authentication time, catches another exception for authentication, replaces Signature CryptoObject with null. This is because next time the app makes an API call, the payload will change, and thus the Signature will be different, so we unfortunately can't rely on the CryptoObject.

### DIFF
--- a/android/src/main/java/io/token/security/AKSSigner.java
+++ b/android/src/main/java/io/token/security/AKSSigner.java
@@ -3,7 +3,6 @@ package io.token.security;
 import android.os.Build;
 import android.security.keystore.KeyPermanentlyInvalidatedException;
 import android.security.keystore.UserNotAuthenticatedException;
-import android.support.v4.hardware.fingerprint.FingerprintManagerCompat;
 import com.google.protobuf.Message;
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/token/security/AKSSigner.java
```diff
@@ -72,8 +72,8 @@ public String sign(String payload) {
             s = Signature.getInstance("SHA256withECDSA");
             s.initSign(((PrivateKeyEntry) entry).getPrivateKey());
 
-            // If we are in an old device ad this is a privileged signer / operation
-            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M && keyLevel != Key.Level.LOW) {
+            // If this is a privileged signer / operation
+            if (keyLevel != Key.Level.LOW) {
 
                 // If user authentication has expired
                 if (System.currentTimeMillis() >= userAuthenticationStore.userAuthenticatedTime()
```
hahrot: optional: userAuthenticationStore should have a `isAuthenticated()` method that implements this logic instead

PR #97, opened 2018-01-17
PL-1189 - adding get balance list api call

VKCR:


lib/src/main/java/io/token/Member.java
```diff
@@ -843,6 +843,32 @@ public Money getCurrentBalance(String accountId, Key.Level keyLevel) {
...
     }
 
+    /**
+     * Looks up current balance for a list of accounts.
+     *
+     * @param accountIds list of account ids
+     * @param keyLevel key level
+     * @return map of account id to current balance
+     */
+    public Map<String, Money> getCurrentBalanceList(
```
ab666: why do you call it getCurrentBalance**List** if you are returning a **Map**?

lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -822,6 +828,150 @@ public Money apply(GetBalanceResponse response) {
...
     }
 
+    /**
+     * Get the current balance of a list of accounts.
+     *
+     * @param accountIdList list of account ids
+     * @param keyLevel key level
+     * @return map of account ids to GetBalanceResponses
+     */
+    public Observable<Map<String, Money>> getCurrentBalanceList(
```
ab666: this is a pretty ugly method :)

lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -822,6 +828,150 @@ public Money apply(GetBalanceResponse response) {
...
+                    public Map<String, Money> apply(GetBalanceListResponse response) {
+                        return response
+                                .getResponseMap()
+                                .entrySet()
+                                .stream()
+                                .filter(new Predicate<Map.Entry<String, GetBalanceResponse>>() {
+                                    @Override
+                                    public boolean test(
+                                            Map.Entry<String, GetBalanceResponse> entry) {
+                                        return entry.getValue().getStatus() == SUCCESSFUL_REQUEST
```
ab666: same here

lib/src/main/java/io/token/Member.java
VKCR: You're right, I'll change everything to ...BalanceMap

lib/src/main/java/io/token/rpc/Client.java
VKCR: No lambda expressions...

ab666:
would require changes based on the service side review.

ab666:
Your commit message ("making changes") is not very descriptive

lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -749,35 +749,25 @@ public Boolean apply(GetDefaultAccountResponse response) {
...
                 .getBalance(GetBalanceRequest
                         .newBuilder()
-                        .setPayload(payload)
-                        .setSignature(Signature
-                                .newBuilder()
-                                .setMemberId(memberId)
-                                .setKeyId(signer.getKeyId())
-                                .setSignature(signer.sign(payload)))
+                        .setAccountId(accountId)
+                        .setNonce(UUID.randomUUID().toString())
```
ab666: Why is it needed? can't we just use the same nonce of the request signature?

VKCR: You mean use the request signature as the nonce?

PR #98, opened 2018-01-18
Pl 1049.linking

gianluca-token:
Implement initiateAccountLinking with browser instance.

Can test using https://github.com/tokenio/android-lib-tmp/tree/testing.


lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -1433,4 +1483,69 @@ private String tokenAction(TokenPayload tokenPayload, Action action) {
                 toJson(tokenPayload),
                 action.name().toLowerCase());
     }
+
+    private Observable<Token> transferTokenExternalAuth(
+            final TokenPayload payload,
+            final CreateTransferTokenResponse response,
+            final BrowserFactory browserFactory) {
+        return Single.create(new SingleOnSubscribe<BankAuthorization>() {
```
ab666: How similar is it to initiateAccountLinking?

gianluca-token: Everything except the last block where we call createTransferToken again is the same

gianluca-token: Hmm maybe it would be better to have a "get bank authorization from browser" method somewhere that takes in the linking url, redirect regex and browser factory.

PR #101, opened 2018-01-23
Pl 1049.payments

gianluca-token:
Implements createTransferToken with browser.

Example Use: Suppose the user wants to make a payment from a CMA9 account, using an instance of browser to authenticate.
1. User adds browserFactory to TransferTokenBuilder using .setBrowserFactory.
2. User calls executeAsync on the TransferTokenBuilder; this will initiate the payment and direct the browser to authenticate.
Note: The user does not need to call createTransferToken a second time after authorization is obtained--this call is made automagically.

Concrete code to test this functionality can be found on the 'testing' branch of android-lib-temp.

### DIFF
--- a/lib/src/main/java/io/token/MemberAsync.java
+++ b/lib/src/main/java/io/token/MemberAsync.java
@@ -765,7 +765,7 @@ public class MemberAsync {
 
     /**
      * Creates a new transfer token from a token payload. If external
-     * authorization is required, will throw exception.
+     * authorization is required, throws TransferTokenException.
      *
      * @param payload transfer token payload
      * @return transfer token returned by the server
--- a/lib/src/main/java/io/token/MemberAsync.java
+++ b/lib/src/main/java/io/token/MemberAsync.java
@@ -764,7 +764,8 @@ public class MemberAsync {
     }
 
     /**
-     * Creates a new transfer token from a token payload.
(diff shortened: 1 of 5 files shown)


lib/src/main/java/io/token/TransferTokenBuilder.java
```diff
@@ -131,6 +134,17 @@ public TransferTokenBuilder setBankAuthorization(BankAuthorization bankAuthoriza
         return this;
     }
 
+    /**
+     * Sets the Browser Factory.
```
lahosken: += something like: If the SDK needs to show a web UI to the user for authorization, it will use this. By default, will use .... .

lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -773,6 +773,19 @@ public TransferTokenBuilder createTransferToken(double amount, String currency)
         return client.createTransferToken(payload);
     }
 
+    /**
```
lahosken: About that createTransferToken(TokenPayload payload) up above (not in your change, so github doesn't let me attach this here commentary to it) that doesn't take a BrowserFactory param... If it needs web-by auth, what happens? If a lovely default browserfactory makes everything work smoothly, then YAY. If not, would be good to += its javadoc comment saying if it needs auth, it throws blah blah blah

PR #102, opened 2018-01-24
Workaround for android 6.0 issue

mariano54:


android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -100,16 +100,21 @@ public Key generateKey(Key.Level keyLevel) {
...
+                    // We can also invalidate the key if the user changes their biometrics
+                    builder = builder
+                            .setInvalidatedByBiometricEnrollment(true)
+                            .setUserAuthenticationRequired(keyLevel != Key.Level.LOW)
+                            .setUserAuthenticationValidityDurationSeconds(
+                                    userAuthenticationStore.authenticationDurationSeconds());
                 }
+
                 kpg = KeyPairGenerator.getInstance(
                         KeyProperties.KEY_ALGORITHM_EC, "AndroidKeyStore");
```
ab666: should it be a constant?

android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -100,16 +100,21 @@ public Key generateKey(Key.Level keyLevel) {
                         getAlias(keyLevel),
                         KeyProperties.PURPOSE_SIGN | KeyProperties.PURPOSE_VERIFY)
                         .setAlgorithmParameterSpec(new ECGenParameterSpec("secp256r1"))
-                        .setDigests(KeyProperties.DIGEST_SHA256)
-                        .setUserAuthenticationRequired(keyLevel != Key.Level.LOW)
-                        .setUserAuthenticationValidityDurationSeconds(
-                            userAuthenticationStore.authenticationDurationSeconds());
+                        .setDigests(KeyProperties.DIGEST_SHA256);
```
lahosken: please update the old comment above, which fools me into thinking this code calls setUserAuthenticationRequired and then I get scared when I see it's going away

mariano54: will do

PR #107, opened 2018-02-16
Pl 1049.oauth

gianluca-token:
EDIT: new Friday 2/23
- Add exception for Sberbank to ensure only Token can link.
- linkAccounts performs entire linking process instead of returning access token

(Older stuff)
- Adds linkBankAccounts methods for new OAuth linking flow.
- Modifies initiateAccountLinking method to return access token instead of bank auth payload.

This branch has become the combination of the OAuth linking flow changes and the browser SDK changes (because they are somewhat mutually dependent). Because the 'browser' branch is outdated and uses the legacy linking flow, I would like to merge this into the 'browser' branch and only then merge the 'browser' branch into master.


lib/src/main/java/io/token/Member.java
```diff
@@ -376,15 +389,47 @@ public Notification getNotification(String notificationId) {
      *
      * @param bankInfo the bank info
      * @param browserFactory the browser factory
-     * @return account linking bank authorization
+     * @return access token
```
lahosken: s/access token/oauth access token/
in context of Member.java , "access token" mostly means Token.io's smart tokens for AISPs.

lib/src/main/java/io/token/Member.java
```diff
@@ -376,15 +389,47 @@ public Notification getNotification(String notificationId) {
...
+            @Override
+            public void subscribe(final SingleEmitter<String> emitter) throws Exception {
+                final Browser browser = browserFactory.create();
+                browser.url()
+                        .filter(new Predicate<URL>() {
+                            @Override
+                            public boolean test(URL url) {
+                                if (url
+                                        .toExternalForm()
+                                        .matches(".*token.io.*#.*token=.+")) {
```
lahosken: should I worry that a bank could use a URL that matches this but isn't in our domain? My regexp-wary instincts are barking at me to do complex things with slashes to ensure domain-ness... but I dunno if that's useful, or if I'm just overly regexp-wary.

lib/src/main/java/io/token/Member.java
```diff
@@ -393,6 +438,7 @@ public Notification getNotification(String notificationId) {
      * @param authorization an authorization to accounts, from the bank
```
lahosken: `@deprecated use BLAH BLAH BLAH instead` (maybe BLAH BLAH BLAH is linkBankAccounts?)

lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -468,6 +471,7 @@ public Notification apply(GetNotificationResponse response) {
      * @param authorization an authorization to accounts, from the bank
      * @return list of linked accounts
```
lahosken: @deprecated use BLAH BLAH BLAH here too

lib/src/main/java/io/token/Member.java
gianluca-token: I don't think we need to worry because the server can't actually access the data in the fragment--that's only available to the browser. But if you have a more precise regex in mind I'm happy to use that instead.

lahosken: how about adding a slash to make `.*token.io/.*#.*token=.+` , to reject token.ionbank.com ?

lib/src/main/java/io/token/Member.java
```diff
@@ -406,7 +406,7 @@ public void subscribe(final SingleEmitter<List<Account>> emitter) throws Excepti
                             public boolean test(URL url) {
                                 if (url
                                         .toExternalForm()
-                                        .matches(".*token.io.*#.*access_token=.+")) {
+                                        .matches(".*token.io/.*#.*access_token=.+")) {
```
ab666: it could be '/', '?', or '#'

ab666: (".*token.io([/?]?.*#).*access_token=.+"

PR #108, opened 2018-02-27
PL 1374

yuhanniu:
GetBanks becomes unauthenticated call

### DIFF
--- a/lib/src/main/java/io/token/TokenIO.java
+++ b/lib/src/main/java/io/token/TokenIO.java
@@ -31,6 +31,7 @@ import io.reactivex.functions.Function;
 import io.token.gradle.TokenVersion;
 import io.token.proto.banklink.Banklink.BankAuthorization;
 import io.token.proto.common.alias.AliasProtos.Alias;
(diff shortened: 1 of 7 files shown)


lib/src/main/java/io/token/TokenIO.java
```diff
@@ -342,6 +344,15 @@ public Member completeRecoveryWithDefaultRule(
                 .blockingSingle();
     }
 
+    /**
+     * Returns a list of all available banks for linking.
```
lahosken: Maybe leave out "all"? I think we (plan to?) hide some un-announced ones

lib/src/main/java/io/token/TokenIOAsync.java
```diff
@@ -408,4 +410,15 @@ public MemberAsync apply(MemberProtos.Member member) throws Exception {
                     }
                 });
     }
+
+
+    /**
+     * Returns a list of all token enabled banks.
```
lahosken: "all"

PR #109, opened 2018-03-01
PL-1382: adding applySCA client method

VKCR:


lahosken:
how/when do we intend/hope sdk users use this?

do we want to automatically call it anytime the user does something involving their thumbprint/whatever, SCA'ing all their accounts?

or just in response to some stepup?

or something?

PR #113, opened 2018-03-12
PL-1395: Allow generate/verify CSRF token from SDK

VKCR:
I don't really know what "check that the nonce hash is in state" means. For now I check if the state contains the nonce hash substring.


lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -83,6 +93,11 @@
  * getMember an existing one and switch to the authenticated {@link Client}.
  */
 public final class UnauthenticatedClient {
+    private static final Alias TOKEN = Alias.newBuilder()
```
colin-token: this should be configurable, though defaulting to Token is ok. Don't we already have this constant somewhere else? (for recovery)

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -527,6 +561,34 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
+    /**
+     * Verify that the state contains the nonce's hash, and that the signature of the payload
+     * (state | tokenId) is valid.
+     *
+     * @param payload request signature payload
+     * @param nonce nonce
+     * @param signature signature
+     * @return completable
+     */
+    public Completable verifyState(
```
colin-token: We should try to abstract away the details as much as possible and provide a wrapper class for the state to allow the user to pass data through without having to think about stuff like nonce/tokenId, etc. The calls should look like just creation + verify/retrieve data from their perspective

VKCR: I don't think I follow. I'm following your lucidchart diagram. I see three different elements: the nonce (the CSRF token), the state, and the signature of the (state|tokenID) payload. Why wrap it?

Also I didn't really create any specific methods for token creation, since it's just hashing a nonce. Should I add a method for that in utilities? I think it'd be a bit overkill.

colin-token:
We want the user of the SDK to be able to create and retrieve their own state. Right now you have generated state, but it doesn't seem like it's used anywhere. Ideally the interface would look something like this:

TokenRequestUrl
-- String crsfToken
-- String Url

TokenRequestUrl createTokenRequestUrl(String tokenRequestId, String state)
String parseTokenRequestState(String callbackUrl, String csrfToken)

The create and parse methods will add the nonce, nonce hash, etc. transparently and use the token if it is provided (we probably want an overloaded method for if the csrfToken is not provided as well)

The best way to work with an SDK is to think about it from the client's usability perspective: you may need to make adjustments to the interface to make it as easy as possible for them to use and abstract away the complexity as much as possible.

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,31 @@ public Member completeRecoveryWithDefaultRule(
...
+    /**
+     * Generate the token request authentication url from a request ID and a state string.
+     *
+     * @param requestId request id
+     * @param state state
+     * @return token request authentication url
+     * @throws MalformedURLException malformed url exception
+     */
+    public TokenRequestGeneratedUrl generateTokenRequestUrl(String requestId, String state)
+            throws MalformedURLException {
```
colin-token: We shouldn't be bubbling the checked exception up like this here. Should catch and rethrow (there's an "exceptions" folder)

lib/src/main/java/io/token/TokenRequestGeneratedUrl.java
```diff
@@ -0,0 +1,67 @@
...
+import static io.token.util.Util.hashString;
+import static java.lang.String.format;
+
+import com.google.auto.value.AutoValue;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+@AutoValue
+public abstract class TokenRequestGeneratedUrl {
```
colin-token: `TokenRequestUrl`?

lib/src/main/java/io/token/TokenRequestGeneratedUrl.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String PROTOCOL = "https";
+    private static final String DOMAIN = "web-app.token.io";
```
colin-token: make domain/path configurable + also vary by environment

lib/src/main/java/io/token/TokenRequestQueryParser.java
```diff
@@ -0,0 +1,79 @@
...
+import io.token.exceptions.InvalidTokenRequestQuery;
+import io.token.proto.ProtoJson;
+import io.token.proto.common.security.SecurityProtos.Signature;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@AutoValue
+public abstract class TokenRequestQueryParser {
+    private static final String TOKEN_ID_HEADER = "token-id";
```
colin-token: these aren't headers, so I'd call them fields instead

lib/src/main/java/io/token/TokenRequestQueryParser.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.auto.value.AutoValue;
+import io.token.exceptions.InvalidTokenRequestQuery;
+import io.token.proto.ProtoJson;
+import io.token.proto.common.security.SecurityProtos.Signature;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@AutoValue
+public abstract class TokenRequestQueryParser {
```
colin-token: package private?

lib/src/main/java/io/token/TokenRequestState.java
```diff
@@ -0,0 +1,53 @@
...
+    public static TokenRequestState create(String nonceHash, String state) {
+        return new AutoValue_TokenRequestState(nonceHash, state);
+    }
+
+    public abstract String getNonceHash();
+
+    public abstract String getState();
+
+    public String toSerializedState() {
+        byte[] data = SerializationUtils.serialize(this);
```
colin-token: let's use gson to do this to keep it language independent

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -527,14 +573,105 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
-                    .build());
+    /**
+     * Generate the token request authentication url from a request ID and a state string.
+     *
+     * @param requestId request id
+     * @param state state
+     * @return token request authentication url
+     * @throws MalformedURLException malformed url exception
+     */
+    public Observable<TokenRequestGeneratedUrl> generateTokenRequestUrl(
```
colin-token: we don't need this to be observable if it's not asynchronous

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -527,14 +573,105 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
+     * Verify that the state contains the nonce's hash, and that the signature of the token request
+     * payload is valid. Return the extracted original state.
+     *
+     * @param tokenRequestUrl token request url
+     * @param nonce nonce
+     * @return the extracted original state
+     */
+    public Observable<String> extractTokenRequestState(URL tokenRequestUrl, String nonce) {
+        final TokenRequestQueryParser parser = TokenRequestQueryParser
+                .create(tokenRequestUrl.getQuery());
```
colin-token: .parse?

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -527,14 +573,105 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
+            String tokenId,
+            String serializedState) {
+        return RequestSignaturePayload.newBuilder()
+                .setTokenId(tokenId)
+                .setState(serializedState)
+                .build();
+    }
+
+    private Member getTokenMember() {
+        String tokenMemberId = getMemberId(TOKEN).blockingSingle();
```
colin-token: these should not be blocking since this interface is async

lib/src/main/java/io/token/util/Util.java
```diff
@@ -211,4 +222,17 @@ public void run() {
...
+    /**
+     * Hashes a string.
+     *
+     * @param value value to hash
+     * @return hash
+     */
+    public static String hashString(String value) {
+        return Hashing
+                .sha256()
+                .hashString(value, Charset.forName("ASCII"))
```
colin-token: verify that this doesn't break url encoding stuff

lib/src/main/java/io/token/TokenRequestGeneratedUrl.java
VKCR: had to move things around a bit but it should be backwards compatible.

VKCR: Didn't make path configurable though, I don't know what the configurations would be.

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +364,29 @@ public Member completeRecoveryWithDefaultRule(
...
+     */
+    public TokenRequestGeneratedUrl generateTokenRequestUrl(String requestId, String state) {
+        return async.generateTokenRequestUrl(requestId, state).blockingSingle();
+    }
+
+    /**
+     * Verify that the state contains the nonce's hash, and that the signature of the token request
+     * payload is valid. Return the extracted original state
+     *
+     * @param tokenRequestUrl token request url
```
mariano54: Also more comment on what the url is

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +364,29 @@ public Member completeRecoveryWithDefaultRule(
         return async.getBanks().blockingSingle();
     }
 
+    /**
+     * Generate the token request authentication url from a request ID and a state string.
```
mariano54: Better comment on what this does

lib/src/main/java/io/token/TokenRequestGeneratedUrl.java
mariano54: Yeah I agree with vivek, we should have nonce in there or they will forget it's there

lib/src/main/java/io/token/TokenRequestQueryParser.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.google.auto.value.AutoValue;
+import io.token.exceptions.InvalidTokenRequestQuery;
+import io.token.proto.ProtoJson;
+import io.token.proto.common.security.SecurityProtos.Signature;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@AutoValue
```
mariano54: Comment on how this is used

lib/src/main/java/io/token/TokenRequestQueryParser.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final String STATE_FIELD = "state";
+    private static final String SIGNATURE_FIELD = "signature";
+
+    /**
+     * Create a new instance of TokenRequestQueryParser.
+     *
+     * @param query query
+     * @return instance of TokenRequestQueryParser
+     */
+    public static TokenRequestQueryParser parse(String query) {
```
mariano54: Comment on how this is used

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -527,14 +569,83 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
+    public Observable<TokenRequestGeneratedUrl> generateTokenRequestUrl(
+            String requestId,
+            String state,
+            TokenIO.TokenCluster tokenCluster) {
+        final TokenRequestGeneratedUrl generatedUrl = TokenRequestGeneratedUrl.create(
+                requestId,
+                state,
+                tokenCluster);
+
+        return Observable.fromCallable(new Callable<TokenRequestGeneratedUrl>() {
```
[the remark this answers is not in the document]
mariano54: We need to return the URL and the nonce right? not just a string

lib/src/main/java/io/token/TokenIO.java
VKCR: done

lib/src/main/java/io/token/TokenRequestQueryParser.java
VKCR: done

lib/src/main/java/io/token/TokenRequestQueryParser.java
VKCR: moved into csrf package

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +364,29 @@ public Member completeRecoveryWithDefaultRule(
...
+
+    /**
+     * Verify that the state contains the nonce's hash, and that the signature of the token request
+     * payload is valid. Return the extracted original state
+     *
+     * @param tokenRequestUrl token request url
+     * @param nonce nonce
+     * @return the extracted original state
+     */
+    public String extractTokenRequestState(URL tokenRequestUrl, String nonce) {
```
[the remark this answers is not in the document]
VKCR: done

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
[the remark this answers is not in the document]
VKCR: correct. Decided to call the object which contains both "CsrfToken", since this is essentially what we are constructing and returning.

lib/src/main/java/io/token/TokenIO.java
VKCR: added a bit more info

lib/src/main/java/io/token/TokenRequestQueryParser.java
VKCR: Added more comments.

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
VKCR: removed from client, but keeping in TokenIOAsync. I have to, to access cluster information.

VKCR:
In the end I left the serialization/deserialization methods in TokenRequestState, for they need to be accessible from the parser.

lib/src/main/java/io/token/csrf/TokenRequestCallbackParameters.java
```diff
@@ -0,0 +1,82 @@
...
+    private static final String TOKEN_ID_FIELD = "token-id";
+    private static final String STATE_FIELD = "state";
+    private static final String SIGNATURE_FIELD = "signature";
+
+    /**
+     * Parses the token request callback URL's parameters. Extracts the state, the token ID, and
+     * the signature over (state | token ID).
+     *
+     * @param query token request callback query
+     * @return TokenRequestCallbackParser instance
```
colin-token: not called parser anymore

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,41 @@ public Member completeRecoveryWithDefaultRule(
...
+     * Generate a CSRF token (a nonce).
+     *
+     * @return CSRF token
+     */
+    public String generateCsrfToken() {
+        return async.generateCsrfToken().blockingSingle();
+    }
+
+    /**
+     * Generate a Token request URL from a request ID, an original state, a CSRF token and a token
```
colin-token: doesn't take in a token cluster does it

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -525,7 +564,8 @@ public TokenIOAsync buildAsync() {
                     cryptoEngine != null
                             ? cryptoEngine
                             : new TokenCryptoEngineFactory(new InMemoryKeyStore()),
-                    devKey);
+                    devKey,
+                    new CsrfTokenManager(tokenCluster));
```
colin-token: Can we pass tokenCluster into TokenIOAsync instead of the manager and instantiate the manager in the async class

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,41 @@ public Member completeRecoveryWithDefaultRule(
         return async.getBanks().blockingSingle();
     }
 
+    /**
+     * Generate a CSRF token (a nonce).
+     *
+     * @return CSRF token
+     */
+    public String generateCsrfToken() {
```
colin-token: can this method be static?

lib/src/main/java/io/token/TokenIOAsync.java
```diff
@@ -432,4 +442,56 @@ public MemberAsync apply(MemberProtos.Member member) throws Exception {
         UnauthenticatedClient unauthenticated = ClientFactory.unauthenticated(channel);
         return unauthenticated.getBanks();
     }
+
+    /**
+     * Generate a CSRF token (a nonce).
+     *
+     * @return CSRF token
+     */
+    public Observable<String> generateCsrfToken() {
```
colin-token: static?

lib/src/main/java/io/token/TokenIOAsync.java
```diff
@@ -432,4 +442,56 @@ public MemberAsync apply(MemberProtos.Member member) throws Exception {
...
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param tokenRequestCallbackUrl token request callback url
+     * @param nonce nonce
+     * @return the extracted original state
+     */
+    public Observable<String> parseTokenRequestCallbackUrl(
+            final URL tokenRequestCallbackUrl,
+            final String nonce) {
```
colin-token: `csrfToken` instead of `nonce` to be consistent

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,41 @@ public Member completeRecoveryWithDefaultRule(
...
+     * @param csrfToken csrf token
+     * @return token request url
+     */
+    public URL generateTokenRequestUrl(String requestId, String state, String csrfToken) {
+        return async.generateTokenRequestUrl(requestId, state, csrfToken).blockingSingle();
+    }
+
+    /**
+     * Parse the token request callback URL to extract the state, the token ID and the signature of
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,41 @@ public Member completeRecoveryWithDefaultRule(
...
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param tokenRequestCallbackUrl token request callback url
+     * @param nonce nonce
+     * @return the extracted original state
+     */
+    public String parseTokenRequestCallbackUrl(URL tokenRequestCallbackUrl, String nonce) {
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -363,6 +365,41 @@ public Member completeRecoveryWithDefaultRule(
...
+        return async.generateTokenRequestUrl(requestId, state, csrfToken).blockingSingle();
+    }
+
+    /**
+     * Parse the token request callback URL to extract the state, the token ID and the signature of
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param tokenRequestCallbackUrl token request callback url
+     * @param nonce nonce
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/TokenIOAsync.java
```diff
@@ -432,4 +442,56 @@ public MemberAsync apply(MemberProtos.Member member) throws Exception {
...
+                csrfToken));
+    }
+
+    /**
+     * Parse the token request callback URL to extract the state, the token ID and the signature of
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param tokenRequestCallbackUrl token request callback url
+     * @param nonce nonce
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/csrf/CsrfTokenManager.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Parse the token request callback URL to extract the state, the token ID and the signature of
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param member member
+     * @param tokenRequestCallbackUrl token request callback URL
+     * @param nonce nonce
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/csrf/CsrfTokenManager.java
```diff
@@ -0,0 +1,154 @@
...
+     * @return original state
+     */
+    public String parseTokenRequestCallbackUrl(
+            Member member,
+            URL tokenRequestCallbackUrl,
+            String nonce) {
```
colin-token: s/nonce/csrfToken

lib/src/main/java/io/token/csrf/CsrfTokenManager.java
```diff
@@ -0,0 +1,154 @@
...
+     * (state | token ID). Verify that the state contains the nonce's hash, and that the signature
+     * of the token request payload is valid. Return the extracted original state.
+     *
+     * @param member member
+     * @param tokenRequestCallbackUrl token request callback URL
+     * @param nonce nonce
+     * @return original state
+     */
+    public String parseTokenRequestCallbackUrl(
+            Member member,
```
colin-token: `trustedMember`? to clarify which member this is

colin-token:
Is the JSON in the state url encoded anywhere? I commented on this earlier, but I don't see where it's done

lib/src/main/java/io/token/TokenIO.java
VKCR: why make it static?

lib/src/main/java/io/token/TokenIO.java
VKCR: fixed

lib/src/main/java/io/token/TokenIO.java
VKCR: done

lib/src/main/java/io/token/TokenIO.java
VKCR: done

lib/src/main/java/io/token/TokenIO.java
VKCR: done, and elsewhere

lib/src/main/java/io/token/TokenIO.java
VKCR: done

lib/src/main/java/io/token/TokenIOAsync.java
VKCR: why make it static?

lib/src/main/java/io/token/TokenIOAsync.java
VKCR: done there and elsewhere

lib/src/main/java/io/token/TokenIOAsync.java
VKCR: done

lib/src/main/java/io/token/csrf/CsrfTokenManager.java
VKCR: I put tokenMember to keep with the other names

lib/src/main/java/io/token/csrf/TokenRequestCallbackParameters.java
VKCR: fixed

VKCR:
Sorry missed that, added encoding of state parameter in URL

lib/src/main/java/io/token/TokenIO.java
colin-token: Because it's just a pure utility function - doesn't modify the state and just generates a token that can be used. Shouldn't need an instance of the manager to use it, even though we are in the current flow

VKCR: Well it's just a wrapper for the generateNonce() utility function. We could get rid of it altogether. Thoughts?

colin-token: yup sounds good - so user just calls manager directly right?

VKCR: yeah, he'd call generateTokenRequestUrl(generateNonce(), state)

VKCR: removing it and merging

PR #121, opened 2018-03-21
PL-1415: request token and add key notification + getTokenIdFromRefID api call

VKCR:


lib/src/main/java/io/token/Member.java
```diff
@@ -1085,6 +1085,27 @@ public Signature requestSignature(String tokenId, String state) {
...
+    }
+
+    /**
+     * Get a transfer token ID from its ref ID.
+     *
+     * @param refId ref id
+     * @param accountId account id
+     * @return token id
+     */
+    public String getTransferTokenIdFromRefId(String refId, String accountId) {
```
ab666: why do you need account-id if you have reference-id?

VKCR: routing. And it will be avoided in the future because we will be polling aggregation service.

lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -776,6 +776,17 @@ public TransferTokenBuilder createTransferToken(double amount, String currency)
...
     }
 
+    /**
+     * Creates a new transfer token from a token payload.
+     *
+     * @param payload transfer token payload
+     * @param tokenRequestId token request id
+     * @return transfer token returned by the server
+     */
+    public Observable<Token> createTransferToken(TokenPayload payload, String tokenRequestId) {
```
ab666: referenceId?

lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -786,6 +797,21 @@ public TransferTokenBuilder createTransferToken(double amount, String currency)
...
+    /**
+     * Creates an access token built from a given {@link AccessTokenBuilder}.
+     *
+     * @param accessTokenBuilder an {@link AccessTokenBuilder} to create access token from
+     * @param tokenRequestId token request id
+     * @return the access token created
+     */
+    public Observable<Token> createAccessToken(
+            AccessTokenBuilder accessTokenBuilder,
+            String tokenRequestId) {
```
ab666: referenceId?

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -391,6 +392,17 @@ public String parseTokenRequestCallbackUrl(URL tokenRequestCallbackUrl, String c
...
     }
 
+    /**
+     * Get a token ID based on a token's tokenRequestId.
+     *
+     * @param tokenRequestId token request id
+     * @param type type of the token
+     * @return token id
+     */
+    public String getTokenId(String tokenRequestId, TokenType type) {
```
ab666: Do we need getTokenId?  Why can't we have "getAccessToken(referenceId)" and "getTransferToken(referenceId)"?

PR #122, opened 2018-03-23
Access token linking

gianluca-token:
- Adds method linkAccounts(bankId, accessToken) in case the user doesn't want to use browser (For AMAIZ, who wants to use the sdk from the server side.)
- Change redirect url to webapp

### DIFF
--- a/lib/src/main/java/io/token/MemberAsync.java
+++ b/lib/src/main/java/io/token/MemberAsync.java
@@ -1162,9 +1162,7 @@ public class MemberAsync {
      * @return web-app callback url
      */
     public String getWebAppCallbackUrl() {
-        return cluster == null
-                ? "http://localhost:5000/auth/callback"
-                : String.format("https://%s/auth/callback", cluster.webAppUrl());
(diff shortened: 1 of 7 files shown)


lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -1152,6 +1156,15 @@ public Money apply(Balance balance) throws Exception {
         return client.getPairedDevices();
     }
 
+    /**
+     * Get the url pointing to web-app callback endpoint.
+     *
+     * @return web-app callback url
+     */
+    public String getWebAppCallbackUrl() {
```
colin-token: do we need this or can we just use the tokencluster directly

gianluca-token: OK tell me whether this is better or worse...

colin-token: If it's only used in one method, can we just construct it at the top? Don't think anyone else is going to use web app callback url right?

PR #130, opened 2018-03-30
PL-1454

gianluca-token:
Right now there is a "createTestBankAccount" sdk method that returns bank auth JSON. Since we are phasing out bank auth JSON, I want to add two methods and deprecate the old one:
- String generateTestBankAccount(balance, currency) // returns access token to link the account
- Account generateLinkedBankAccount(balance, currency) // actually links the account

### DIFF
--- a/lib/src/main/java/io/token/Member.java
+++ b/lib/src/main/java/io/token/Member.java
@@ -30,6 +30,7 @@ import io.reactivex.functions.Function;
 import io.token.exceptions.BankAuthorizationRequiredException;
 import io.token.proto.PagedList;
 import io.token.proto.banklink.Banklink.BankAuthorization;
+import io.token.proto.banklink.Banklink.OauthBankAuthorization;
 import io.token.proto.common.alias.AliasProtos.Alias;
 import io.token.proto.common.bank.BankProtos.BankInfo;
 import io.token.proto.common.blob.BlobProtos.Attachment;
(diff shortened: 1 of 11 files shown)


lib/src/main/java/io/token/Member.java
```diff
@@ -1032,6 +1035,34 @@ public Blob getTokenBlob(String tokenId, String blobId) {
...
     }
 
+    /**
+     * Generates a test bank account in a fake bank and links the account.
+     *
+     * @param balance account balance to set
+     * @param currency currency code, e.g. "EUR"
+     * @return the linked account
+     */
+    public Account generateLinkedBankAccount(double balance, String currency) {
```
ab666: change the old instead

lib/src/main/java/io/token/Member.java
```diff
@@ -971,14 +972,31 @@ public Blob getTokenBlob(String tokenId, String blobId) {
...
     }
 
+    /**
+     * Creates a test bank account in a fake bank and links the account.
+     *
+     * @param balance account balance to set
+     * @param currency currency code, e.g. "EUR"
+     * @return the linked account
+     */
+    public Account createLinkedBankAccount(double balance, String currency) {
```
ab666: the name is confusing... we might add a feature to create bank accounts in the future.
the name of the method should clearly indicate the account is fake and to be used for testing only.

gianluca-token: createLinkedTestAccount?

gianluca-token: or createLinkedTestBankAccount?

PR #135, opened 2018-04-11
Add performance testing env

drobinson123:


mariano54:
Will this be a general purpose env? If so, we will need to add config in all of the services

aktoken:
@mariano54 it's only for load/performance testing and tuning. Configs are already in place.

PR #137, opened 2018-04-13
implement new getBanks method

sibinlu:


yuhanniu:
All fields are optional. We should have a "super method" and create some useful overrides for it. `redeemToken` is a good reference: https://github.com/tokenio/sdk-java/blob/master/lib/src/main/java/io/token/MemberAsync.java#L1026

PR #145, opened 2018-05-01
add set/get receipt email methods for member

gianluca-token:
fix aar bug

### DIFF
--- a/lib/src/main/java/io/token/rpc/Client.java
+++ b/lib/src/main/java/io/token/rpc/Client.java
@@ -1279,7 +1279,7 @@ public final class Client {
     /**
      * Replaces member's receipt contact.
      *
-     * @param memberId member ID of member whose email we will set
+     * @param memberId member ID of member whose receipt contact we will set
      * @param contact receipt contact to set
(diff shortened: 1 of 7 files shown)


lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -1272,6 +1276,41 @@ public Blob apply(GetProfilePictureResponse response) {
...
+     */
+    public Observable<ReceiptContact> getReceiptContact(String memberId) {
+        return Util
+                .toObservable(gateway.getReceiptContact(GetReceiptContactRequest.newBuilder()
+                        .setMemberId(memberId)
+                        .build()))
+                .map(new Function<GetReceiptContactResponse, ReceiptContact>() {
+                    @Override
+                    public ReceiptContact apply(
+                            GetReceiptContactResponse getReceiptContactResponse) {
```
yuhanniu: nit: it seems we don't need newline here?

gianluca-token: without it, the line is 103 chars long

PR #146, opened 2018-05-01
PL-1542: add deleteMember call

VKCR:


VKCR:
changing branch to reflect ticket

PR #148, opened 2018-05-02
PL-1583: fix null pointer exception bug

VKCR:


lib/src/main/java/io/token/MemberAsync.java
```diff
@@ -171,7 +171,7 @@ public String apply(MemberProtos.Member member) throws Exception {
                 .map(new Function<List<Alias>, Alias>() {
                     public Alias apply(List<Alias> aliases) throws Exception {
                         if (aliases.isEmpty()) {
-                            return null;
+                            return Alias.getDefaultInstance();
```
sam-token: I'm not sure we want to introduce sentinels into the public api. Since we don't have Optional available, maybe something with Rx Maybe or empty Observables would work, if not we may just have to treat it as an error condition and throw a proper exception.

VKCR: I decided to throw exceptions. Note that this causes a bit of code duplication in TransferTokenBuilder.

lib/src/main/java/io/token/exceptions/NoAliasesFoundException.java
```diff
@@ -0,0 +1,10 @@
+package io.token.exceptions;
+
+import static java.lang.String.format;
+
+public class NoAliasesFoundException extends RuntimeException {
+    public NoAliasesFoundException(String memberId) {
+        super(format("Member could not be resolved for alias %s", memberId));
```
sam-token: s/alias/memberId

lib/src/main/java/io/token/TokenIO.java
```diff
@@ -142,7 +141,7 @@ public boolean aliasExists(Alias alias) {
      * Looks up member id for a given alias.
      *
      * @param alias alias to check
-     * @return member id if alias already exists, null otherwise
+     * @return member id
```
sam-token: Maybe add a note that it throws if not found. For this method and firstAlias

lib/src/main/java/io/token/exceptions/MemberNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package io.token.exceptions;
+
+import static java.lang.String.format;
+
+import io.token.proto.ProtoJson;
+import io.token.proto.common.alias.AliasProtos.Alias;
+
+public class MemberNotFoundException extends RuntimeException {
```
sam-token: maybe extend StatusRuntimeException w/ Status.NOT_FOUND

PR #151, opened 2018-05-08
PL-1588: add realms where relevent

VKCR:


lib/src/main/java/io/token/TokenIO.java
```diff
@@ -307,6 +427,17 @@ public String beginRecovery(Alias alias) {
...
     }
 
+    /**
+     * Begins account recovery.
+     *
+     * @param alias the alias used to recover
+     * @param realm realm of the alias
+     * @return the verification id
+     */
+    public String beginRecovery(Alias alias, String realm ) {
```
sam-token: nit: rm space after realm

lib/src/main/java/io/token/TransferTokenBuilder.java
```diff
@@ -306,6 +306,18 @@ public TransferTokenBuilder setToAlias(Alias toAlias) {
         return this;
     }
 
+    /**
+     * Sets the realm of the payee.
+     *
+     * @param realm realm
+     * @return builder
+     */
+    public TransferTokenBuilder setToRealm(String realm ) {
```
sam-token: nit: rm space after realm

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -612,7 +627,7 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
      * @return token member
      */
     public Observable<Member> getTokenMember() {
-        return getMemberId(TOKEN).flatMap(
+        return getMemberId(TOKEN, TOKEN_REALM).flatMap(
```
sam-token: `TOKEN_REALM` is used here, empty string elsewhere. Any reason for the differences?

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -105,13 +106,15 @@ public UnauthenticatedClient(GatewayServiceFutureStub gateway) {
      * Checks if a given alias already exists.
      *
      * @param alias alias to check
+     * @param realm realm of the alias
      * @return {@code true} if alias already exists, {@code false} otherwise
      */
-    public Observable<Boolean> aliasExists(Alias alias) {
+    public Observable<Boolean> aliasExists(Alias alias, String realm) {
```
sam-token: should all the methods in this class be overloaded as well?

lib/src/main/java/io/token/TokenIO.java
VKCR: fixd

lib/src/main/java/io/token/TransferTokenBuilder.java
VKCR: fixd

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
VKCR: Doesn't seem necessary to me, the entry point for developers is tokenIO and tokenIOasync right?

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
VKCR: Yes, this method is used to query the "token.io" member which we know inhabits the token realm. However for the other calls I don't think we should make assumptions if no realm is given. Of course right now in practice the result would be the same since we default to the token realm if no realm is provided, but maybe this will change in the future.

sam-token: I see, makes sense.

lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
sam-token: Good point, should be fine

PR #161, opened 2018-05-21
PL-1519: add provider parameter to getBanks

VKCR:


lib/src/main/java/io/token/TokenIO.java
```diff
@@ -562,13 +562,40 @@ public Member completeRecoveryWithDefaultRule(
      *     Defaults to name if not specified.
      * @return a list of banks
      */
+    @Deprecated
```
sam-token: instead of deprecating, maybe just add a single overloaded method with all fields?

VKCR: This would lead to a multiplication of methods. As I see it there are three methods: one which accepts only bank IDs, one which accepts filter parameters, and one which accepts everything. It makes sense for the method qhich accepts filter parameters to accept the provider, which is why I'm deprecating this one and adding a new one with the provider parameter added.

sam-token: As-is right now this is still a breaking change, you'd need to add one more method (getBanks with everything except provider)

VKCR: Oh right. Added.

PR #171, opened 2018-06-01
can generate key with expiration date using CryptoEngine

gianluca-token:


lib/src/main/java/io/token/security/TokenCryptoEngine.java
```diff
@@ -99,4 +110,14 @@ private Key toPublicKey(SecretKey secretKey) {
                 .setPublicKey(crypto.serialize(secretKey.getPublicKey()))
                 .build();
     }
+
+    private Key toPublicKey(SecretKey secretKey, long expirationMs) {
+        return Key.newBuilder()
```
yuhanniu: nit: `toPublicKey(secretKey).toBuilder().setExpiresAtMs(expirationMs).build()`

PR #172, opened 2018-06-01
Only set artifactor username/password if environment variable is present

colin-token:


android/build.gradle
```diff
@@ -51,8 +51,10 @@ artifactory {
             } else {
                 repoKey = 'public-libs-release-local'
             }
-            username = "${artifactory_user}"     // The publisher user name
-            password = "${artifactory_password}" // The publisher password
+            if (System.env['artifactory_user'] != null) {
```
sibinlu: we should check for System.env['artifactory_password'] !=null for some edge cases

PR #179, opened 2018-06-19
Update default sample test dev key

sam-token:


colin-token:
Do we need a version bump?

sam-token:
Since it's just sample changes we shouldn't need to bump

PR #180, opened 2018-06-20
Pl 164.keystore

gianluca-token:
For pentesting, we enabled generating keys with expiration dates from the sdk. The KeyStore needs to know which keys are expired so it doesn't return a signer for an expired key, especially when a valid key exists that it could have returned instead. So I am adding clock to the KeyStore using it to filter out expired keys.

The Android crypto engine uses Android API's keystore. Newer versions of the API support key expiration nicely, but for older versions, I've needed to hack-ily encode the expiration date into the "alias" (the identifier for the keypair in the Android keystore). I've separated the logic so that we can just remove the chunks of code that support older versions of the API later on.

NOTE: I made a small change to AKSVerifier because it wasn't working when I tested it at first and I didn't see any uses of it, even in tests. If there are usages, my change will probably break them, so please let me know if this is the case.

### DIFF
--- a/android/src/main/java/io/token/security/AKSCryptoEngine.java
+++ b/android/src/main/java/io/token/security/AKSCryptoEngine.java
@@ -29,7 +29,9 @@ import android.security.KeyPairGeneratorSpec;
 import android.security.keystore.KeyGenParameterSpec;
 import android.security.keystore.KeyInfo;
 import android.security.keystore.KeyProperties;
+
 import com.google.common.hash.Hashing;
+
 import io.token.exceptions.SecureHardwareKeystoreRequiredException;
 import io.token.proto.common.security.SecurityProtos.Key;
 import io.token.util.codec.ByteEncoding;
@@ -51,6 +53,9 @@ import java.util.Calendar;
 import java.util.Enumeration;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
 import javax.security.auth.x500.X500Principal;
 
 /**
@@ -71,9 +76,9 @@ public final class AKSCryptoEngine implements CryptoEngine {
      * Creates an instance. If useSecureHardwareKeystoreOnly is true and insecure keystore is
      * detected, a SecureHardwareKeystoreRequiredException error will be thrown.
      *
-     * @param memberId member ID
-     * @param context context, to draw UI
-     * @param userAuthenticationStore stores the last time the user authenticated
+     * @param memberId                      member ID
+     * @param context                       context, to draw UI
+     * @param userAuthenticationStore       stores the last time the user authenticated
      * @param useSecureHardwareKeystoreOnly true if use secure hardware keystore only
      */
     public AKSCryptoEngine(
@@ -102,7 +107,7 @@ public final class AKSCryptoEngine implements CryptoEngine {
      */
     @Override
     public Key generateKey(Key.Level keyLevel) {
-        String serializedPk = generatePublicKey(keyLevel);
+        String serializedPk = generatePublicKey(keyLevel, null);
         return Key.newBuilder()
                 .setId(keyIdFor(serializedPk))
                 .setAlgorithm(KEY_ALGORITHM)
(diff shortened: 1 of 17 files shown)


lib/src/main/java/io/token/util/SystemClock.java
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright (c) 2017 Token, Inc.
```
yuhanniu: same

lib/src/test/java/io/token/util/TestClock.java
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) 2017 Token, Inc.
```
yuhanniu: same

lib/src/test/java/io/token/util/TestClock.java
```diff
@@ -0,0 +1,34 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package io.token.util;
+
+public class TestClock implements Clock {
+    public static final long FUTURE = 2000000000000L;
+    public static final long PAST = 1000000000000L;
+    private long time = 1500000000000L;
```
yuhanniu: I think it would be better if we give this test clock one or two extra methods to tweak it time comparatively to the real time, instead of hardcoding the time here.

lib/src/main/java/io/token/util/Clock.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2017 Token, Inc.
```
yuhanniu: nit: 2018

android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -269,10 +269,12 @@ public Verifier createVerifier(String keyId) {
      * @return alias
      */
     private String createAlias(Key.Level keyLevel, @Nullable Long expirationMs) {
-        return KEY_NAME + "-" + memberId + "-" + keyLevel.toString() + "-"
-                + (Build.VERSION.SDK_INT < Build.VERSION_CODES.M && expirationMs != null
-                ? expirationMs
-                : UUID.randomUUID().toString().replace("-", ""));
+        String alias = KEY_NAME + "-" + memberId + "-" + keyLevel.toString() + "-"
```
ab666: String.format()

lib/src/main/java/io/token/security/TokenCryptoEngine.java
```diff
@@ -103,17 +104,14 @@ public Verifier createVerifier(String keyId) {
     }
 
     private Key toPublicKey(SecretKey secretKey) {
+        long expirationMs = secretKey.getExpirationMs() == null
+                ? 0
+                : secretKey.getExpirationMs();
```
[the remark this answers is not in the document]
sibinlu: Not sure if ExpirationMs is the best naming. I remember we used ExpiresAtMs everywhere in the proto. Maybe it is better to be consistent if you consider it necessary too.

android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -237,13 +260,24 @@ public Verifier createVerifier(String keyId) {
...
      * @return alias
      */
-    private String getAlias(Key.Level keyLevel) {
-        return KEY_NAME + "-" + memberId + "-" + keyLevel.toString();
+    private String createAlias(Key.Level keyLevel, @Nullable Long expirationMs) {
+        String alias = String.format("%s-%s-%s-%s",
+                KEY_NAME, memberId, keyLevel.toString(),
+                UUID.randomUUID().toString().replace("-", ""));
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M && expirationMs != null) {
+            alias = alias + "-" + expirationMs;
```
sibinlu: Having different key store alias for different version could be a bad idea. Assuming the users upgrade their system from the old version to the new one. The keys might be lost due to key alias mismatch.

android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -254,12 +288,26 @@ private String getAlias(Key.Level keyLevel) {
...
         return Key.Level.valueOf(aliasParts[2]);
     }
 
+    /**
+     * Gets the expiration date of the key associated with the alias.
+     *
+     * @param alias key store alias
+     * @return expiration date in milliseconds
+     */
+    private static Long getExpiration(String alias) {
```
sibinlu: nowhere says this function is only for the old key store aliases which is confusing.

android/src/main/java/io/token/security/AKSCryptoEngine.java
gianluca-token: Keys generated in the old version should be fine in the new version. They will have the pattern keyName-memberId-keyLevel so they will be returned from getAliasPrefix(), just as the new key keyName-memberId-keyLevel-UUID will. I need to change the aliases so that I can have two keys at the same level for the same member--right now the aliases would be duplicated.

lib/src/main/java/io/token/security/TokenCryptoEngine.java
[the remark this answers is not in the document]
gianluca-token: I slightly prefer using null for absent values. It's unlikely, but maybe someone wants to test the behavior of expired keys by putting 0 in for the expiration.

PR #182, opened 2018-06-27
PL-1706: add realms to alias.

VKCR:


lib/src/main/java/io/token/util/Util.java
```diff
@@ -180,7 +176,8 @@ public static String hashAlias(Alias alias) {
         if (alias.getType() == USERNAME) {
             return alias.getValue();
         }
-        return hashAndSerializeJson(alias);
+
+        return hashAndSerializeJson(alias.toBuilder().clearRealm().build());
```
sam-token: use proto function?

build.gradle
```diff
@@ -24,7 +24,7 @@ buildscript {
 allprojects {
     ext {
         ver = [
-            tokenProto           : '1.0.493',
+            tokenProto           : '1.0.495',
```
sam-token: version

PR #183, opened 2018-06-28
use get default agent instead of hard code

sibinlu:


sibinlu:
will merge after the change is sandbox

PR #188, opened 2018-07-24
Expose proto account objects

yuhanniu:
The `Account` wrapper we have in the SDK doesn't expose all the information that the proto account object has, ie, `AccountFeatures`, `AccountTags`, etc. We would need to update `Account` every time we add something to the proto account object to keep them synced. I am thinking we should expose the proto account object in the `Account` wrapper, so in the wrapper interface we only have frequently/commonly used attributes.

A more immediate use case if that we need to update the merchant-proxy to support data access functionalities, which uses the Java SDK inside. We would need the Java SDK to expose the proto account object so that the merchant-proxy can access all the account information.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,5 +33,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '1.0.150'
+    version = '1.0.151'
 }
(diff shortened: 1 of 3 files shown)


sibinlu:
imo, we should either have a wrapper for the features/tags or drop this sdk level account structure.
it is weird to have something like `account.protoAccount.supportPayment` in code.

PR #189, opened 2018-07-27
implment new realm logic in sdk

sibinlu:
1 Update to the newest proto
2 Implement new realm logic in sdk
3 Update hash function

### DIFF
--- a/lib/src/main/java/io/token/util/Util.java
+++ b/lib/src/main/java/io/token/util/Util.java
@@ -75,17 +75,18 @@ import javax.annotation.Nullable;
 public abstract class Util {
 
     /**
+     * The token realm.
+     */
+    public static final String TOKEN_REALM = "token";
+
+    /**
      * The token alias.
      */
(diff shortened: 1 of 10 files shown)


lib/src/main/java/io/token/util/Util.java
```diff
@@ -158,32 +159,28 @@ public static Alias normalizeAlias(Alias rawAlias) {
...
      */
-    public static MemberOperation toAddAliasOperation(Alias alias, String realm) {
+    public static MemberOperation toAddAliasOperation(Alias alias) {
         return MemberOperation
                 .newBuilder()
                 .setAddAlias(MemberAliasOperation
                         .newBuilder()
-                        .setAliasHash(hashAlias(alias))
-                        .setRealm(realm))
+                        .setAliasHash(hashAlias(alias)))
```
VKCR: MemberAliasOperation still needs the realm (since hashing the alias removes that information).

PR #191, opened 2018-08-01
RD-491: updated dependencies

aktoken:
also changed order of repositories in android module cause in some cases build fails because gradle unable to find dependency in jcenter and doesn't try to search in goolge repo

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
 allprojects {
     ext {
         ver = [
-            tokenProto           : '1.0.516',
+            tokenProto           : '1.0.520',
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -24,14 +24,14 @@ buildscript {
 allprojects {
     ext {
         ver = [
-            tokenProto           : '1.0.515',
-            tokenRpc             : '1.0.122',
-            tokenSecurity        : '1.0.40',
-            tokenUtil            : '1.0.16',
+            tokenProto           : '1.0.516',
```
sam-token: use latest proto?

PR #193, opened 2018-08-02
Ab.java

ab666:


lib/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -692,12 +692,12 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
      */
     public Observable<String> getTokenId(String tokenRequestId) {
         return toObservable(gateway
-                .getTokenId(GetTokenIdRequest.newBuilder()
+                .getTokenRequestResult(GetTokenRequestResultRequest.newBuilder()
```
sam-token: I think this isn't implemented in the gateway yet. I believe @sibinlu is working on that.

PR #194, opened 2018-08-07
onBehalfOf is always included in the metadata when set

yuhanniu:


yuhanniu:
PTAL

yuhanniu:
Solved by https://github.com/tokenio/sdk-java/pull/195

PR #195, opened 2018-08-09
RD-804

yuhanniu:
`Member.actAs()` returns a `DelegateMember`
`MemberAsync.actAs()` returns a `DelegateMemberAsync`

I think this wrapper approach logically makes more sense than the interface approach. It would also work with the Javascript SDK so wouldn't need to deal with Javascript not having class hierarchy.

Named the new member as `DelegateMember` because it would still make sense if we decided to add more power to access tokens later.

### DIFF
--- a/lib/src/main/java/io/token/DelegateMember.java
+++ b/lib/src/main/java/io/token/DelegateMember.java
@@ -1,0 +1,125 @@
+/**
+ * Copyright (c) 2018 Token, Inc.
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * <p>
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * <p>
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package io.token;
+
+import io.token.proto.PagedList;
+import io.token.proto.common.member.MemberProtos.AddressRecord;
+import io.token.proto.common.security.SecurityProtos.Key;
+import io.token.proto.common.transaction.TransactionProtos.Balance;
+import io.token.proto.common.transaction.TransactionProtos.Transaction;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Acts as another member by using a granted access token.
+ */
+public class DelegateMember {
+    private final Member member;
+
+    public DelegateMember(Member member, String tokenId) {
+        this.member = member;
+        this.member.useAccessToken(tokenId);
+    }
+
+    /**
+     * Looks up member addresses.
+     *
+     * @return a list of addresses
+     */
+    public List<AddressRecord> getAddresses() {
+        return member.getAddresses();
+    }
+
+    /**
+     * Looks up an address by id.
+     *
+     * @param addressId the address id
+     * @return an address record
+     */
+    public AddressRecord getAddress(String addressId) {
+        return member.getAddress(addressId);
+    }
+
+    /**
+     * Looks up funding bank accounts linked to Token.
+     *
+     * @return list of linked accounts
+     */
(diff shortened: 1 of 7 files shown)


lib/src/main/java/io/token/DelegateMember.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param accountId the account id
+     * @param transactionId ID of the transaction
+     * @param keyLevel key level
+     * @return transaction
+     */
+    public Transaction getTransaction(
+            String accountId,
+            String transactionId,
+            Key.Level keyLevel) {
+        return getTransaction(accountId, transactionId, keyLevel);
```
sam-token: `member.getTransaction`?

lib/src/main/java/io/token/Member.java
```diff
@@ -165,6 +165,16 @@ public void clearAccessToken() {
...
     }
 
+    /**
+     * Act as another member using an granted access token.
+     *
+     * @param tokenId the token id
+     * @return a delegate member that acts as the grantor of the access token
+     */
+    public DelegateMember actAs(String tokenId) {
+        return new DelegateMember(this.clone(), tokenId);
```
sam-token: this approach looks fine to me, but I'm curious why not use the interface? I'd image it'd save you the effort of forwarding all the calls to member and all you'd have to do here is 
```
DelegateMember member = this.clone(); // Maybe a more applicable interface name
member.useAccessToken(id);
return member;
```

yuhanniu: * I think this wrapper approach logically makes more sense.
Otherwise we have `Member` implements `DelegateMember` while `Member` isn't a `DelegateMember` by business logic.
* It would also work with the Javascript SDK. Javascript doesn't have interface
* With the interface approach people can still cast `DelegateMember` to `Member`, the wrapper approach can prevent that.

yuhanniu:
@sibinlu Could you review please? We'll need to do the same thing to the objc sdk so it is better we get on the same page.
The idea is to make the usage of the access tokens thread-safe while keeping the `impersonation` design.

lib/src/main/java/io/token/Member.java
sam-token: I think if we have the correct naming it makes sense. What JS does is not really relevant, it's a different language with different idioms. I don't think casting is an issue; it'd have to be intentional and even then it's not like it allows one to do anything malicious.

I do prefer the interface approach, I think it will be easier to maintain; but I think this approach works too and won't be to different from the client perspective.

yuhanniu: Naming: What we need is not a basic version of the `Member` object but a `Member` object that has the all the methods that can be used with an access token, so I don't think a name like `LimitedMember` is as good as `DelegateMember`.
JS: Implementation consistency at least doesn't hurt if it is not a plus.
Casting: People can still mess with their own accounts after casting thinking they are operating on someone else's accounts. It is not like it can happen often and it is not malicious but why don't we shut down the opportunities of making that kind of mistakes when we can? 
In terms of maintenance cost, the difference is only one line of code per method.

yuhanniu:
@sibinlu Have moved `clone()` and made `DelegateMemebr` and `DelegateMemberAsync`'s constructers package private.
The idea is that when you use an access token, you behave exactly like the member that granted the access token, you pretend to be someone else.

yuhanniu:
@ab666 I have stated the reason why I don't think interface is a good idea. A `Member` is not necessarily a `Representable`, it could be just itself not representing anyone, having `Member` implements `Representable` looks weird to me. @sam-token probably likes your idea. @sibinlu what do you think?

ab666:
@yuhanniu Representable means that the Member can be represented by someone else.  In fact, all members are representable by means of access tokens.  You can think of the access token as a power fo attorney for another member to act as an agent on behalf of the member they represent.

yuhanniu:
@ab666 ok that makes sense. `Representable` doesn't sound very intuitive to me tho, but it is not like I have a better name. If we go with `Representable`, should we name the method `represent`? ie
```java
Representable otherMember = member.represent(tokenId);
```

ab666:
Nah, no need to call the method represent, "forAccessToken" is fine in my opinion.  Better name anyone, @sam-token @colin-token ?

yuhanniu:
Updated. Any other suggestions regarding naming or anything before merging? @all

lib/src/main/java/io/token/Member.java
```diff
@@ -154,17 +155,43 @@ public void useAccessToken(String accessTokenId) {
...
+    /**
+     * Creates a {@link Representable} that acts as another member using the access token
+     * that was granted by that member.
+     *
+     * @param tokenId the token id
+     * @param customerInitiated whether the call is initiated by the customer
+     * @return the {@link Representable}
+     */
+    public Representable forAccessToken(String tokenId, boolean customerInitiated) {
+        MemberAsync cloned = (MemberAsync) async.forAccessToken(tokenId, customerInitiated);
```
sam-token: maybe include a package-private variant of async.forAccessToken that returns a MemberAsync so you don't have to rely on the cast here?

yuhanniu: Is the cast not reliable under certain circumstances?

sam-token: As-is, not really. But it does circumvent the type system and relies on an implementation detail in a separate file. If, for some reason, we were ever to change the concrete type of MemberAsync.forAccessToken to something that can't be cast to MemberAsync, the compiler wouldn't tell us we need to update Member.forAccessToken too.

yuhanniu: Makes sense. Now I'll have to come up with another name :slightly_smiling_face:

yuhanniu:
I think the difficulty on the naming is inherited from the design. Still not a fan of the design, I feel like we are just trying to be smart :joy:

colin-token:
I'm also not convinced that returning just a `Member` is a bad idea - I thought we were going to send the access token on every request and let gateway handle whether the access is allowed or not

yuhanniu:
There are local calls like getMemberId. Also the idea is to make the `impersonation` more intuitive with a good naming.

PR #199, opened 2018-08-14
keys() fetches from server

gianluca-token:


build.gradle
```diff
@@ -33,5 +33,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '1.1.3'
```
yuhanniu: why 5?

gianluca-token: I have another PR here that I was planning to merge before this one

PR #201, opened 2018-08-22
RD-255: add createClaimedMember method, implement TPP license related api changes

VKCR:


lib/src/main/java/io/token/TokenRequest.java
```diff
@@ -51,24 +52,101 @@ public String getName() {
...
+            this.tokenPayload = tokenPayload;
+            this.options = new HashMap<>();
+        }
+
+        public Builder setOption(TokenRequestOptions option, String value) {
+            options.put(option.getName(), value);
+            return this;
+        }
+
+        public Builder setOptions(Map<String, String> options) {
```
sam-token: nit: call it `putAllOptions`?
`setOptions` may imply you're replacing existing options map with the given map, while `putAll` makes the intent more clear

VKCR: decided to call it addAllOptions, and also replaced setOption with addOption.

PR #202, opened 2018-09-10
Allow member to be created with specific id

colin-token:


lib/src/main/java/io/token/TokenIO.java
```diff
@@ -191,6 +190,24 @@ public Member createMember(Alias alias) {
...
+     * gives more control over the member creation process.
+     *
+     * <p>Adds an alias and a set of auto-generated keys to the member.</p>
+     *
+     * @param alias nullable member alias to use, must be unique. If null, then no alias will
+     *     be created with the member
+     * @param memberId member id
+     * @return newly created member
+     */
+    public Member createMember(final Alias alias, final String memberId) {
```
sam-token: is it still `createMember` if the already exists? Would `initializeTestMember` or something similar more more sense?

colin-token:
We need a way of creating an instance of `Member` with a member ID that we already have. We can theoretically copy the member creation code to SDK tests, but it might make it easy to accidentally get out of sync/be much harder to update

lib/src/main/java/io/token/TokenIO.java
colin-token: I can change it to `setupMember` since it's taking in a member ID that has already been created, but it's not necessarily a test member at this point so calling it `test` would be misleading.

sam-token:
do we need to copy the logic or can we just do
```
Member member = TokenIO.getMember(memberId);
member.addAlias(alias);
member.approveKeys(low, standard, privileged);
```
or is there more setup work that's missing?
I'm concerned that adding a method to the public interface that is essentially testing-only may make the api unnecessarily confusing.

PR #212, opened 2018-10-04
RD-1064: Add/Remove trusted beneficiaries

yuhanniu:


lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -1647,6 +1652,60 @@ public Completable verifyAffiliate(String memberId) {
...
     }
 
+    /**
+     * Adds a trusted beneficiary for whom the SCA will be skipped.
+     *
+     * @param payload the payload of the request
+     * @return a completable
+     */
+    public Completable addTrustedBeneficiary(TrustedBeneficiary.Payload payload) {
+        Signer signer = crypto.createSigner(PRIVILEGED);
```
sibinlu: Standard makes more sense. Here we should use the same key as endorse since they have the same purpose. Enjoy.

yuhanniu: I enjoyed it very much, thank you.

PR #214, opened 2018-10-04
RD-1129: s/redeemer/to for transfer tokens

VKCR:


VKCR:
The pollNotificationSampleTest broke when I made the change, the reason being that the redeemer of the token was set as the payer, and the payer redeemed the token. Changed it to payee. I suppose a consequence of this change is the payer cannot be the redeemer anymore?

sam-token:
@VKCR yes it is going to be a different notification

PR #221, opened 2018-10-25
remove deprecated login methods

yuhanniu:
It has been more than 10 months since the methods were deprecated.


yuhanniu:
@grace09thompson has kindly communicated with our relatively old customer xcordis to make sure that they are not using the login method, merging.

PR #223, opened 2018-10-25
use updated AddKey proto

gianluca-token:


samples/src/main/java/io/token/sample/ProvisionDeviceSample.java
```diff
@@ -23,7 +25,12 @@ public static Key provisionDevice(TokenIO tokenIO, Alias alias) {
             return k.getLevel() == Key.Level.LOW;
         }).findFirst().orElse(null);
         // ask user (on "regular" device) to approve one of our keys
-        NotifyStatus status = tokenIO.notifyAddKey(alias, "SDK Sample", lowKey);
+        NotifyStatus status = tokenIO.notifyAddKey(
```
yuhanniu: please make sure this doesn't break documentation.

gianluca-token: Well it doesn't "break" it but only the first line will be visible--good catch. I will make a slate PR to fix.

PR #226, opened 2018-11-02
RD-1366: remove wildcard access token resources

gianluca-token:
https://github.com/tokenio/slate/pull/332


yuhanniu:
`forAllAccounts()` should be deprecated too?
I am not sure which way is better - maybe you should finish the new token request flow first? Otherwise it seems like you would need to modify the sdk, samples and tests again later(for those which are related to token request flow).

PR #229, opened 2018-11-21
Rd 1454

gianluca-token:
- Update TokenRequest
- Add updateTokenRequest method
- Add notifyCreateAndEndorseToken method


lib/src/main/java/io/token/TokenRequest.java
```diff
@@ -54,77 +56,139 @@ public String getName() {
...
+    @Deprecated
     @Nullable public abstract String getUserRefId();
 
+    @Deprecated
     @Nullable public abstract String getCustomizationId();
 
     @Nullable public abstract Customization getCustomization();
 
+    @Deprecated
     public String getOption(TokenRequestOptions option) {
```
yuhanniu: also deprecate this `TokenRequestOptions` enum?

lib/src/main/java/io/token/TokenRequest.java
```diff
@@ -141,12 +205,23 @@ public Builder setCustomization(Customization customization) {
          * @return TokenRequest instance
          */
         public TokenRequest build() {
-            return new AutoValue_TokenRequest(
-                    tokenPayload,
-                    options,
-                    userRefId,
-                    customizationId,
-                    customization);
+            return requestPayload == null
```
yuhanniu: same

PR #232, opened 2018-12-11
RD-1568: Security related metadata

sam-token:
- Allows users to supply security related metadata
- Makes AuthenticationContext stateless. Previously it used thread-local storage which can cause issues for Members that are shared across threads. The stateful nature of AuthenticationContext also allowed potential race conditions for asynchronous calls (eg first call sets keys level, begins connection, second call sets new key level, first call receives onMessage() event in authentication interceptor and reads incorrect key level). Now a new AuthenticationContext is created for each rpc call

### DIFF
--- a/lib/src/main/java/io/token/rpc/Client.java
+++ b/lib/src/main/java/io/token/rpc/Client.java
@@ -25,6 +25,7 @@ package io.token.rpc;
 import static com.google.common.base.Strings.nullToEmpty;
 import static io.token.proto.ProtoJson.toJson;
 import static io.token.proto.banklink.Banklink.AccountLinkingStatus.FAILURE_BANK_AUTHORIZATION_REQUIRED;
+import static io.token.proto.common.security.SecurityProtos.Key.Level.LOW;
 import static io.token.proto.common.security.SecurityProtos.Key.Level.PRIVILEGED;
 import static io.token.proto.common.security.SecurityProtos.Key.Level.STANDARD;
 import static io.token.proto.common.token.TokenProtos.TokenSignature.Action.CANCELLED;
@@ -67,9 +68,9 @@ import io.token.proto.common.notification.NotificationProtos.Notification;
 import io.token.proto.common.notification.NotificationProtos.NotifyStatus;
 import io.token.proto.common.notification.NotificationProtos.StepUp;
 import io.token.proto.common.security.SecurityProtos.Key;
+import io.token.proto.common.security.SecurityProtos.SecurityMetadata;
 import io.token.proto.common.security.SecurityProtos.Signature;
 import io.token.proto.common.subscriber.SubscriberProtos.Subscriber;
-import io.token.proto.common.token.TokenProtos;
 import io.token.proto.common.token.TokenProtos.Token;
 import io.token.proto.common.token.TokenProtos.TokenOperationResult;
 import io.token.proto.common.token.TokenProtos.TokenPayload;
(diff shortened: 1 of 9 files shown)


lib/src/main/java/io/token/rpc/Client.java
```diff
@@ -1818,20 +1903,12 @@ public int hashCode() {
         return (memberId + (onBehalfOf == null ? "" : onBehalfOf)).hashCode();
     }
 
-    @Override
-    public Client clone() {
-        return new Client(memberId, crypto, gateway);
+    private AuthenticationContext authenticationContext() {
```
yuhanniu: why don't we just store a mutable AuthenticationContext or an `AuthenticationContext.Builder` in the client?

lib/src/main/java/io/token/rpc/ClientFactory.java
```diff
@@ -58,16 +61,24 @@ public static UnauthenticatedClient unauthenticated(ManagedChannel channel) {
...
-                        new ErrorHandlerFactory()
-                )
-        );
-        return new Client(memberId, crypto, stub);
+                        new ErrorHandlerFactory()));
+
+        GatewayProvider provider = new GatewayProvider() {
+            @Override
+            public GatewayServiceFutureStub withAuthentication(AuthenticationContext context) {
+                return stub.withInterceptors(new Interceptor(new ClientAuthenticatorFactory(
```
yuhanniu: not sure, does this clear existing intercepters?

sam-token: No, it just returns a new stub with the given interceptors added to the channel. The existing interceptors are not affected

lib/src/main/java/io/token/rpc/Client.java
sam-token: In a multithreaded environment modifying the builder (or mutating the context) could cause race conditions. We wouldn't be able to just do `gateway.withAuthentication(builder.setKeyLevel(STANDARD).build())` because this may affect another concurrent request. We'd need to do something like `gateway.WithAuthentication(builder.build().toBuilder().setKeyLevel().build())`. It's simpler and easy to reason about if the authentication context is stateless

sam-token: Yeah the last example would be effectively the same. As long as all request-specific fields are added the the final call it would work. The only difference would be style and I think I'd prefer the stateless object. There's also less of a chance of accidentally introducing a race condition later.

lib/src/main/java/io/token/rpc/ClientFactory.java
yuhanniu: I don't think we need the factory anymore

sam-token: The factory is required by token's Interceptor to convert SimpleInterceptor to a ClientInterceptor

lib/src/main/java/io/token/rpc/Client.java
yuhanniu: I think the benefit would be it looks cleaner especially when we add more fields to the AuthenticationContext later, we would just have a list of stuff that belong to the AuthenticationContext that we cache in the client

yuhanniu:
can you also make a ticket to make change csharp sdk implementation too? It currently has the same implmentation

lib/src/main/java/io/token/rpc/Client.java
sam-token: I'd argue these belong in the client. The authentication context is just a wrapper of fields we pass to the interceptor. From a style perspective `withAuthentication(authenticationContext(keyLevel, other new fields...))` seems cleaner than `withAuthentication(context.build(keyLevel, other new fields...)`

sam-token:
Yeah RD-1568 has all SDKs as subtasks

PR #235, opened 2018-12-13
RD-1700

yuhanniu:


yuhanniu:
on second thought, maybe we should just remove them if our mobile app is doing replace and endorse in two steps. No one else uses it. I see there is an action [here](https://tokenio.atlassian.net/browse/RD-1645), once we get confirmation from @sibinlu we should be able to just remove the method.

PR #236, opened 2018-12-13
RD-1505: Combine async and sync methods into the same class

gianluca-token:
NOTE: I have not updated the samples and tests to use the new code. I wanted to get feedback on the directory structure before doing this. I will update the samples and tests before merging this PR.


yuhanniu:
It feels like when you have an "api" package, a lot of the other stuff should belong to it, like the classes under "tokenrequest" and "exceptions", so I am not sure about "api".

I would suggest that we do as less changes as possible instead of deprecating so many classes, to avoid the trouble to update the all the documents and the hassle for the sdk users.

I recognize the difficulty to move Account and Member to different packages without major refactoring. Sorry I didn't realize this earlier, maybe we should combine the ticket with the modularization thing? Hopefully we would do some re-structuring and thus make things easier.

gianluca-token:
Maybe it's best to combine this with the modularization changes? I don't think it makes sense to move the biggest classes into a subpackage but leave the smaller classes in io.token.

yuhanniu:
sounds good

PR #242, opened 2019-01-23
RD-1804: TPP SDK

gianluca-token:
TPP SDK implementation. Trying to merge into the branch with common + user.

Note: Most things were copy-pasted from lib/ but I significantly changed TokenRequest in order to provide a better user experience and to match the other SDKs better. So feedback on TokenRequest.java in particular would be appreciated :)


tpp/src/main/java/io/token/tpp/tokenrequest/TokenRequest.java
```diff
@@ -0,0 +1,313 @@
...
+
+package io.token.tpp.tokenrequest;
+
+import com.google.auto.value.AutoValue;
+import io.token.proto.common.alias.AliasProtos;
+import io.token.proto.common.alias.AliasProtos.Alias;
+import io.token.proto.common.member.MemberProtos.Customization;
+import io.token.proto.common.token.TokenProtos;
+import io.token.proto.common.token.TokenProtos.ActingAs;
+import io.token.proto.common.token.TokenProtos.TokenMember;
```
yuhanniu: unused protos

tpp/src/main/java/io/token/tpp/tokenrequest/TokenRequest.java
```diff
@@ -0,0 +1,313 @@
...
+
+    public abstract TokenRequestOptions getTokenRequestOptions();
+
+    /**
+     * Create a new Builder instance for an access token request.
+     *
+     * @param resources access token resources
+     * @return Builder instance
+     */
+    public static Builder accessTokenRequestBuilder(ResourceType... resources) {
```
yuhanniu: I suggest we have two separate builder classes, e.g. `AccessTokenRequestBuilder` and `TransferTokenRequestBuilder`, which can both extend an internal `Builder` class that controls the common part aka `TokenRequestOptions`. This way people won't be able to make mistakes and you won't need do the `!requestPayload.hasTransferBody()` kind of checks. And we probably won't need to do put arguments in the constructors of the builders - everything to be set via setters.

yuhanniu:
1. testCompile user module in the tests to create tokens / create and link test accounts
2. consider removing `createAndLinkTestBankAccountImpl`

yuhanniu:
TokenRequest.Builder can be private
Can remove Destinations class

gianluca-token:
TokenRequest.Builder needs to be public so that the setters are visible.

PR #243, opened 2019-01-24
RD-1851 + RD-1805

gianluca-token:
Add bank and partner modules. Not sure what to do about bank samples due to inability to create tokens--suggestions appreciated.


yuhanniu:
Why do we need ability to create tokens? Do we actually run the samples in integration tests?

question: do we know if the getTransfers/getTokens are actually implemented on the server side? i.e. if a bank calls getTransfers/getTokens does it get all transfers/tokens created by its users?

Also do we need to createAndLinkTestAccount methods in those SDKs?

yuhanniu:
discussed offline: shouldn't publish bank sdk before backend support is implemented

PR #246, opened 2019-02-07
RD-1964: move some TokenRequest logic to core so that user can use it

gianluca-token:


core/src/main/java/io/token/TokenClient.java
```diff
@@ -580,6 +584,72 @@ public MemberRecoveryOperation getRecoveryAuthorizationBlocking(
...
+    }
+
+    /**
+     * Updates an existing token request.
+     *
+     * @param requestId token request ID
+     * @param options new token request options
+     * @return completable
+     */
+    public Completable updateTokenRequest(String requestId, TokenRequestOptions options) {
```
yuhanniu: Do TPPs need to update the token request?

yuhanniu:
missing version change?

build.gradle
```diff
@@ -32,5 +32,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '2.0.0-Beta'
+    version = '2.0.1-Beta'
```
yuhanniu: `2.0.0-Beta2`?

tpp/src/main/java/io/token/tpp/tokenrequest/TokenRequestCallbackParameters.java
```diff
@@ -30,6 +30,7 @@
 import io.token.tpp.exceptions.InvalidTokenRequestQuery;
 import io.token.tpp.util.Util;
 
+import java.security.KeyPair;
```
yuhanniu: remove

core/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -27,10 +27,12 @@
 import static io.token.proto.common.security.SecurityProtos.Key.Level.LOW;
 import static io.token.proto.common.security.SecurityProtos.Key.Level.PRIVILEGED;
 import static io.token.proto.common.security.SecurityProtos.Key.Level.STANDARD;
+import static io.token.rpc.util.Converters.toCompletable;
 import static io.token.util.Util.generateNonce;
 import static io.token.util.Util.normalizeAlias;
 import static io.token.util.Util.toObservable;
 
+import io.reactivex.Completable;
```
yuhanniu: remove unused imports

tpp/src/main/java/io/token/tpp/rpc/UnauthenticatedClient.java
```diff
@@ -93,6 +93,21 @@ public Member apply(GetMemberResponse response) {
...
     }
 
+    /**
+     * Updates an existing token request.
+     *
+     * @param requestId token request ID
+     * @param options new token request options
+     * @return completable
+     */
+    public Completable updateTokenRequest(String requestId, TokenRequestOptions options) {
```
yuhanniu: Can you remind me why do TPPs need to update the token requests?

gianluca-token: ah I deleted it but forgot to commit. There's not a use case for it now; if there is, we can add it later.

build.gradle
gianluca-token: changed it to 2.0.0-beta-2, a la https://mvnrepository.com/artifact/junit/junit

PR #247, opened 2019-02-08
Adjustments for bank sdk

yuhanniu:
Some adjustments to make the implementation to enable bank sdk to reuse the code here.


build.gradle
```diff
@@ -32,5 +32,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '2.0.0-Beta'
+    version = '2.0.0-Beta2'
```
gianluca-token: maybe not important but I found precedence for the pattern 2.0.0-beta-2 here https://mvnrepository.com/artifact/junit/junit.

core/src/main/java/io/token/TokenClient.java
```diff
@@ -208,19 +207,19 @@ public String getMemberIdBlocking(Alias alias) {
      * @param memberId member id
      * @return newly created member
      */
-    protected Observable<Member> setUpMemberImpl(final Alias alias, final String memberId) {
+    protected Observable<Member> setUpMemberImpl(
```
gianluca-token: What if you also add setUpMemberImpl(alias, memberId) that just calls this method with generateKeys() and make generateKeys private?

yuhanniu:
needs changed, closing

PR #248, opened 2019-02-08
RD-1931: add TPP account

gianluca-token:
- Account objects return subclass member
- Make core a Token gradle project.

The member() method inherited by Account subclasses returns core member, so it needs to be overridden.

EDIT: also:
- remove memberId param from getDefault account
- move triggerBalance/Txn to Tpp
- rm token attachments
- add getNotificationBlocking
- add some redeemToken convenience methods
- getMember uses getMemberImpl


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -220,12 +237,12 @@ public Account getAccountBlocking(String accountId) {
      * @param data file data
      * @return attachment
      */
-    public Observable<Attachment> createAttachment(
+    public Observable<Attachment> createBlob(
```
yuhanniu: createBlob returns an Attachment? Should we make those align?

gianluca-token: Discussed with @zixunjian and @sibinlu, we can remove this since we are not supporting token attachments for now and the other uses for blobs (logos, profile pictures) have their own separate calls.

PR #249, opened 2019-02-12
add name and app name to customization

sibinlu:


gianluca-token:
Can you edit the TPP SDK please?

tpp/src/main/java/io/token/tpp/rpc/Client.java
```diff
@@ -286,18 +286,24 @@ public String apply(StoreTokenRequestResponse storeTokenRequestResponse) {
...
      * @param consentText consent text
+     * @param name display name
+     * @param appName corresponding app name
      * @return customization id
      */
     public Observable<String> createCustomization(
             Blob.Payload logo,
             Map<String, String> colors,
-            String consentText) {
+            String consentText,
```
yuhanniu: all the fields here should be nullable, otherwise devs will have to use empty string to me absence.

PR #253, opened 2019-02-21
add android user dependency

gianluca-token:


sibinlu:
Do you want to bump the versions as well?

sibinlu:
drop lib from the Android SDK.

PR #254, opened 2019-02-24
RD-2053: include actor in token-sdk header

gianluca-token:


gianluca-token:
2/27 rebased and added small unrelated change to StoreAndRetrieveTokenRequestSample.java.

PR #255, opened 2019-02-26
RD-2057: add processTokenRequestCallback

gianluca-token:
and fix TokenRequest state setter

### DIFF
--- a/core/src/main/java/io/token/tokenrequest/TokenRequestState.java
+++ b/core/src/main/java/io/token/tokenrequest/TokenRequestState.java
@@ -20,13 +20,18 @@
  * THE SOFTWARE.
  */
 
-package io.token.tpp.tokenrequest;
+package io.token.tokenrequest;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.auto.value.AutoValue;
 import com.google.gson.Gson;
+import io.token.util.codec.ByteEncoding;
 
 @AutoValue
 public abstract class TokenRequestState {
+    private static final Gson gson = new Gson();
+
     public static TokenRequestState create(String csrfTokenHash, String state) {
         return new AutoValue_TokenRequestState(csrfTokenHash, state);
     }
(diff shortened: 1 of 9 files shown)


tpp/src/main/java/io/token/tpp/TokenClient.java
```diff
@@ -377,15 +380,70 @@ public String generateTokenRequestUrlBlocking(
...
+    }
+
+    /**
+     * Processes the token request callback, checking the CSRF token.
+     *
+     * @param callbackParams callback parameter map
+     * @param csrfToken CSRF token
+     * @return TokenRequestCallback object containing the token ID and the original state
+     */
+    public Observable<TokenRequestCallback> processTokenRequestCallback(
```
yuhanniu: I think we should call it `parseTokenRequestCallback` for consistency

core/src/main/java/io/token/tokenrequest/TokenRequestState.java
```diff
@@ -52,7 +58,7 @@ public static TokenRequestState parse(String serialized) {
      * @return serialized state
      */
     public String serialize() {
-        Gson gson = new Gson();
-        return gson.toJson(this);
+        String json = gson.toJson(this);
+        return ByteEncoding.serialize(json.getBytes(UTF_8));
```
yuhanniu: why are we changing the behavior of this call?

tpp/src/main/java/io/token/tpp/TokenClient.java
gianluca-token: The other one is called parseTokenRequestCallbackUrl. If we want to overload (which I like), how about naming both methods `processTokenRequestCallback`?

core/src/main/java/io/token/tokenrequest/TokenRequestState.java
gianluca-token: So that callback_state isn't naked json.

tpp/src/main/java/io/token/tpp/TokenClient.java
gianluca-token: I changed to `parseTokenRequestCallbackParams`

core/src/main/java/io/token/tokenrequest/TokenRequestState.java
yuhanniu: I personally don't mind if the state is a naked json or not, but if we are doing this we should create a ticket to unify the behavior of all the sdks.

PR #256, opened 2019-02-27
fix checkstyle

gianluca-token:


gianluca-token:
this commit both:
(1) makes the sample work without csrf token, and
(2) adds csrf token to the sample (because we encourage clients to use it)

PR #259, opened 2019-03-02
RD-2121: Add support for feature codes

gianluca-token:
Adds `addFeatureCode`/`clearFeatureCode` methods to Member.java and TokenClient.java.


gianluca-token:
Decided on different approach.

PR #262, opened 2019-03-12
RD-2185: add confirmFunds call and access token resource

gianluca-token:


core/src/main/java/io/token/rpc/Client.java
```diff
@@ -397,6 +398,28 @@ public Transaction apply(GetTransactionResponse response) {
...
     }
 
+    /**
+     * Confirm that the given account has sufficient funds to cover the charge.
+     *
+     * @param accountId account ID
+     * @param amount charge amount
+     * @return true if the account's balance is higher than the amount
+     */
+    public Observable<Boolean> confirmFunds(String accountId, MoneyProtos.Money amount) {
```
yuhanniu: nit: import money

core/src/main/java/io/token/rpc/Client.java
```diff
@@ -397,6 +398,28 @@ public Transaction apply(GetTransactionResponse response) {
...
+        return toObservable(gateway
+                .withAuthentication(onBehalfOf())
+                .confirmFunds(ConfirmFundsRequest.newBuilder()
+                        .setAccountId(accountId)
+                        .setAmount(amount)
+                        .build()))
+                .map(new Function<Gateway.ConfirmFundsResponse, Boolean>() {
```
yuhanniu: import

user/src/main/java/io/token/user/AccessTokenBuilder.java
```diff
@@ -187,6 +187,23 @@ public AccessTokenBuilder forTransferDestinations(String accountId) {
...
+     * a given charge.
+     *
+     * @param accountId account ID
+     * @return {@link AccessTokenBuilder}
+     */
+    public AccessTokenBuilder forFundsConfirmation(String accountId) {
+        payload
+                .getAccessBuilder()
+                .addResources(Resource.newBuilder()
+                        .setFundsConfirmation(Resource.FundsConfirmation.newBuilder()
```
yuhanniu: import

core/src/main/java/io/token/Member.java
```diff
@@ -718,6 +718,33 @@ public Balance getBalanceBlocking(String accountId, Key.Level keyLevel) {
...
+    /**
+     * Confirm that the given account has sufficient funds to cover the charge.
+     *
+     * @param accountId account ID
+     * @param amount charge amount
+     * @param currency charge currency
+     * @return true if the account's balance is higher than the amount
+     */
+    public Observable<Boolean> confirmFunds(String accountId, double amount, String currency) {
+        return client.confirmFunds(accountId, MoneyProtos.Money.newBuilder()
```
yuhanniu: import

core/src/main/java/io/token/Member.java
```diff
@@ -718,6 +718,33 @@ public Balance getBalanceBlocking(String accountId, Key.Level keyLevel) {
...
 
+    /**
+     * Confirm that the given account has sufficient funds to cover the charge.
+     *
+     * @param accountId account ID
+     * @param amount charge amount
+     * @param currency charge currency
+     * @return true if the account's balance is higher than the amount
+     */
+    public Observable<Boolean> confirmFunds(String accountId, double amount, String currency) {
```
sibinlu: we should not use float type to present a decimal amount. We should use a decimal type or at least a string.

core/src/main/java/io/token/Member.java
```diff
@@ -718,6 +718,33 @@ public Balance getBalanceBlocking(String accountId, Key.Level keyLevel) {
...
+                .build());
+    }
+
+    /**
+     * Confirm that the given account has sufficient funds to cover the charge.
+     *
+     * @param accountId account ID
+     * @param amount charge amount
+     * @param currency charge currency
+     * @return true if the account's balance is higher than the amount
```
sibinlu: I think `if the account has sufficient funds to cover the charge.` is more accurate. The account's balance might be higher than the amount but the account is frozen.

core/src/main/java/io/token/Member.java
gianluca-token: why should it be different than other places where we use double, like createTransferToken?

sibinlu: They are the same. If that is the case, I think `createTransferToken` is wrong as well. We might want to address this problem in another pr.

gianluca-token: What is wrong with double? It is precise to the second decimal at least.

sibinlu: https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

gianluca-token: Seems ok because we immediately convert it to a string? We don't perform computations on the double, we just provide it as a convenient input then convert it to a more precise type.

yuhanniu: @sibinlu We do convert amounts to BigDecimals in our services when calculations and conversions are needed

sibinlu: I think it is the same reason why we use amount as string in `Money` message and all GRPC design. Grpc don't support decimal very well, so we designed a `Money` message for it.

sibinlu: I'm ok that we can address it later. But IMO if the SDK user treats the currency correctly, double is not convenient to use even it will be converted back to string immediately.

sibinlu: If we have decide to get rid of the double amount, we can use string or Money directly here.

PR #263, opened 2019-03-20
RD-2208/2136: Normalize alias on add, remove

gianluca-token:
- Add MemberMethodsSample
- Add setDefaultAccount

### DIFF
--- a/user-sample/src/main/java/io/token/sample/MemberMethodsSample.java
+++ b/user-sample/src/main/java/io/token/sample/MemberMethodsSample.java
@@ -6,6 +6,7 @@ import static io.token.proto.common.security.SecurityProtos.Key.Level.PRIVILEGED
 import static io.token.proto.common.security.SecurityProtos.Key.Level.STANDARD;
 import static io.token.util.Util.generateNonce;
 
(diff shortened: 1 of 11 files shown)


gianluca-token:
Couple things in one PR here:

@kanyewesternfront You can just look at the first commit which adds MemberMethods sample. I did it a little differently than js--let me know what you think about it. In js, adding, removing and resolving are all in the same method, but here I've split them up so that you can make a separate code snippet for adding/removing before the subsection on resolving.

@sibinlu I have also added setDefaultAccount to Member in this PR.

user-sample/src/main/java/io/token/sample/MemberMethodsSample.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Adds and removes keys.
+     *
+     * @param crypto crypto engine
+     * @param member member
+     */
+    public static void keys(CryptoEngine crypto, Member member) {
+        SecurityProtos.Key lowKey = crypto.generateKey(LOW);
```
yuhanniu: import?

PR #271, opened 2019-04-11
RD-1407: add prepareToken, createToken, signTokenPayload

gianluca-token:
Samples will be added in a separate PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,5 +32,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '2.0.4-beta-0'
+    version = '2.0.4-beta-1'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
 allprojects {
     ext {
         ver = [
-            tokenProto           : '1.1.75',
+            tokenProto           : '1.1.76',
             tokenRpc             : '1.1.25',
             tokenSecurity        : '1.1.6',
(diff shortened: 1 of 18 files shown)


user/src/main/java/io/token/user/rpc/Client.java
```diff
@@ -312,6 +317,59 @@ public Transfer apply(GetTransferResponse response) {
...
+    public Observable<Token> createToken(
+            TokenPayload payload,
+            @Nullable String tokenRequestId,
+            List<Signature> signatures) {
+        CreateTokenRequest.Builder request = CreateTokenRequest.newBuilder()
+                .setPayload(payload);
+        if (tokenRequestId != null) {
+            request.setTokenRequestId(tokenRequestId);
+        }
+        if (!signatures.isEmpty()) {
```
yuhanniu: nit: this check is probably not necessary

build.gradle
```diff
@@ -24,13 +24,13 @@ buildscript {
...
+            tokenProto           : '1.1.76',
             tokenRpc             : '1.1.25',
             tokenSecurity        : '1.1.6',
             rxjava               : '2.1.1',
         ]
     }
 
     group = 'io.token.sdk'
-    version = '2.0.3'
+    version = '2.0.4-beta-0'
```
yuhanniu: is "beta-0" a convention?

gianluca-token: will change to -beta-1

PR #274, opened 2019-04-19
fix ServiceSpecificException

sibinlu:
For some undocumented mystery reasons, Android team decides to mark KeyStore.getEntry as a service specific call. Therefore, whenever get entry is invoked on Android KeyStores, a ServiceSpecificException warning would be thrown.

other fixes:
1. fix android tests. it is not working.
2. clean up imports, in Android recommended order.

### DIFF
--- a/android/src/main/java/io/token/security/AKSCryptoEngine.java
+++ b/android/src/main/java/io/token/security/AKSCryptoEngine.java
@@ -29,10 +29,8 @@ import android.security.KeyPairGeneratorSpec;
 import android.security.keystore.KeyGenParameterSpec;
 import android.security.keystore.KeyInfo;
 import android.security.keystore.KeyProperties;
+
 import com.google.common.hash.Hashing;
-import io.token.exceptions.SecureHardwareKeystoreRequiredException;
-import io.token.proto.common.security.SecurityProtos.Key;
-import io.token.util.codec.ByteEncoding;
 
 import java.io.IOException;
 import java.math.BigInteger;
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -197,74 +215,64 @@ public Signer createSigner(Key.Level keyLevel) {
...
     /**
-     * Gets a key Entry from the keystore, by the keyLevel.
+     * Looks up an unexpired alias in the keystore, by the keyLevel.
      *
      * @param keyLevel keyLevel
-     * @return Entry
+     * @return alias
      */
-    private KeyStore.Entry getKeyFromKeyStore(Key.Level keyLevel) {
+    private String lookupUnexpiredAliasInKeyStore(Level keyLevel) {
```
gianluca-token: nit: lookUp

android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -197,74 +215,64 @@ public Signer createSigner(Key.Level keyLevel) {
...
     /**
-     * Gets a key Entry from the keystore, by the keyId.
+     * Looks up an unexpired certificate in the keystore, by the keyId.
      *
      * @param keyId keyId
-     * @return Entry
+     * @return certificate
      */
-    private KeyStore.Entry getKeyFromKeyStore(String keyId) {
+    private Certificate lookupUnexpiredCertificateInKeyStore(String keyId) {
```
gianluca-token: nit: lookUp

PR #280, opened 2019-05-02
Rd 2379

Prateek-Gupta1:


core/src/main/java/io/token/util/Util.java
```diff
@@ -117,7 +117,8 @@ public static MemberOperation toAddAliasOperation(Alias alias) {
                 .setAddAlias(MemberAliasOperation
                         .newBuilder()
                         .setAliasHash(normalizeAndHash(alias))
-                        .setRealm(alias.getRealm()))
+                        .setRealm(alias.getRealm())
+                        .build())
```
yuhanniu: Inner `builder()` is not needed.

core/src/main/java/io/token/util/Util.java
```diff
@@ -132,7 +133,8 @@ public static MemberOperationMetadata toAddAliasOperationMetadata(Alias alias) {
         return MemberOperationMetadata.newBuilder()
                 .setAddAliasMetadata(AddAliasMetadata.newBuilder()
                         .setAlias(normalized)
-                        .setAliasHash(hash(normalized)))
+                        .setAliasHash(hash(normalized))
+                        .build())
```
yuhanniu: same

PR #281, opened 2019-05-09
added provision to add user specified recovery agent in User sdk.

Prateek-Gupta1:


core/src/main/java/io/token/TokenClient.java
```diff
@@ -189,13 +190,14 @@ public String getMemberIdBlocking(Alias alias) {
     protected Observable<Member> createMemberImpl(
             final Alias alias,
             final CreateMemberType memberType,
-            @Nullable String partnerId) {
+            @Nullable String partnerId,
+            final Optional<String> recoveryAgent) {
```
yuhanniu: why don't we make it nullable to be consistent with the partnerId? Also the SDK in java 7, `Optional` is not available.

core/src/main/java/io/token/TokenClient.java
```diff
@@ -215,9 +217,11 @@ public String getMemberIdBlocking(Alias alias) {
      * @param memberId member id
      * @return newly created member
      */
-    protected Observable<Member> setUpMemberImpl(final Alias alias, final String memberId) {
+    protected Observable<Member> setUpMemberImpl(final Alias alias,
+                                                 final String memberId,
+                                                 final Optional<String> agent) {
         final UnauthenticatedClient unauthenticated = ClientFactory.unauthenticated(channel);
-        return unauthenticated.getDefaultAgent()
+        return unauthenticated.getAgent(agent)
```
yuhanniu: `agent` should be the member id of the recovery agent, not the alias value.
Remove getAgent.

```
return (agent == null ? unauthenticated.getDefaultAgent() : Observable.just(agent))
    .flatMap(...)
```

core/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -459,6 +461,38 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
...
+     * @return the recovery agent id.
+     */
+    public Observable<String> getAgent(Optional<String> agent) {
+        // TODO(sibin): Use GetDefaultAgentRequest instead after the call is available.
+        return agent.map(new java.util.function.Function<String, Observable<String>>() {
+                @Override
+                public Observable<String> apply(String agent) {
+                    return toObservable(gateway.resolveAlias(ResolveAliasRequest.newBuilder()
+                            .setAlias(Alias.newBuilder()
+                                    .setType(Alias.Type.BANK)
```
yuhanniu: An agent can be any member, doesn't need to be a bank.

user/src/main/java/io/token/user/TokenClient.java
```diff
@@ -128,7 +130,35 @@ public static TokenClient create(TokenCluster cluster, String developerKey) {
...
+
+    /**
+     * Creates a new Token member with a set of auto-generated keys, an alias, member type
+     * and recovery agent.
+     *
+     * @param alias nullable member alias to use, must be unique. If null, then no alias will
+     *     be created with the member.
+     * @return newly created member
+     */
+    public Observable<Member> createMember(final Alias alias, final String recoveryAgent) {
```
yuhanniu: make recoveryAgent nullable, so `createMember(final Alias alias)` can delegate to this `createMember`.

PR #302, opened 2019-06-08
Update docs link

nikanka-token:


kanyewesternfront:
Sorry, I checked the link and thought it was something else, not the old docs site. I was under the impression that had already been removed so it confused me. Sorry for missing that the first time.

PR #309, opened 2019-06-24
RD-2309: konsentus integration

VKCR:


core/src/main/java/io/token/Member.java
```diff
@@ -441,6 +442,24 @@ public Completable verifyAlias(String verificationId, String code) {
...
+     * An eIDAS-type alias containing auth number of the TPP should be added to the
+     * member before making this call. The member must be under the realm of a bank.
+     *
+     * @param payload payload containing the member id and the certificate in PEM format
+     * @param signature the payload signed with a private key corresponding to the certificate
+     *     public key using "SHA1withDSA" algorithm
+     *     // TODO: include the algorithm in the payload?
+     * @return a status of the verification process
+     */
+    public Observable<EidasProtos.VerifyEidasStatus> verifyEidas(
```
yuhanniu: This should be in the tpp sdk right?

core/src/main/java/io/token/Member.java
```diff
@@ -441,6 +442,24 @@ public Completable verifyAlias(String verificationId, String code) {
...
+    /**
+     * Verifies eIDAS alias with an eIDAS certificate, containing auth number equal to the value
+     * of the alias.
+     * An eIDAS-type alias containing auth number of the TPP should be added to the
+     * member before making this call. The member must be under the realm of a bank.
+     *
+     * @param payload payload containing the member id and the certificate in PEM format
+     * @param signature the payload signed with a private key corresponding to the certificate
+     *     public key using "SHA1withDSA" algorithm
+     *     // TODO: include the algorithm in the payload?
```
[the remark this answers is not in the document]
VKCR: rm'd

core/src/main/java/io/token/Member.java
VKCR: moved

PR #310, opened 2019-06-24
RD-2283: request BALANCES and TRANSACTIONS access on linking

gianluca-token:


user/src/main/java/io/token/user/TransferTokenBuilder.java
```diff
@@ -124,12 +124,19 @@ public TransferTokenBuilder(Member member, TokenRequest tokenRequest) {
                                 .getTransferBody()
                                 .getCurrency())
                         .setAmount(tokenRequest.getRequestPayload().getTransferBody().getAmount())
-                        .setInstructions(TransferInstructions.newBuilder()
-                                .addAllDestinations(tokenRequest
-                                        .getRequestPayload()
+                        .setInstructions(tokenRequest.getRequestPayload()
+                                .getTransferBody()
+                                .hasInstructions()
+                                ? tokenRequest.getRequestPayload()
```
sibinlu: better to have a variable for tokenRequest.getRequestPayload().getTransferBody() to make the code more clear

PR #311, opened 2019-06-25
update the endpoint for createTransferTokenBuilder

sibinlu:
1. rename to createTransferTokenBuilder which makes more sense.
2. Add endpoint for createTransferTokenBuilder(TokenPayload)
3. From information is optional in TokenRequest and TokenPayload theoretically. If it doesn't present we can fill it with the member.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,5 +32,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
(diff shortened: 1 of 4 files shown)


sibinlu:
Add a commit for app callback url. It could be a different pr but I would like to take the same ride here.

PR #314, opened 2019-07-10
added function to allow addition of source endpoint info in Transfer …

Prateek-Gupta1:
…token.


gianluca-token:
This is part of RD-2885

PR #320, opened 2019-07-22
RD-2221

gianluca-token:
java 8, prepareRecurringTransferToken, redeemRecurringTransferToken, getRecurringTransfer, getRecurringTransfers, standing orders, recurring transfer token request


core/src/main/java/io/token/Account.java
```diff
@@ -203,6 +204,62 @@ public Transaction getTransactionBlocking(String transactionId, Key.Level keyLev
...
     }
 
+    /**
+     * Looks up an existing standing order for a given account.
+     *
+     * @param standingOrderId ID of the standing order
+     * @param keyLevel key level
+     * @return standing order record
+     */
+    public Observable<StandingOrder> getStandingOrder(
```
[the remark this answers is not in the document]
gianluca-token: IMO Key.Level is more descriptive--it's what we have in other places throughout the SDK

user/src/main/java/io/token/user/Member.java
```diff
@@ -957,6 +1090,10 @@ public Token getTokenBlocking(String tokenId) {
             @Nullable String description,
             @Nullable TransferDestination destination,
             @Nullable String refId) {
+        if (!token.getPayload().hasTransfer()) {
+            throw new IllegalArgumentException("Expected transfer token; found "
```
yuhanniu: extra `;`

user/src/main/java/io/token/user/StandingOrderTokenBuilder.java
```diff
@@ -0,0 +1,376 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is used to build a transfer token. The required parameters are member, amount (which
+ * is the lifetime amount of the token), and currency. One source of funds must be set: either
+ * accountId or BankAuthorization. Finally, a redeemer must be set, specified by either alias
+ * or memberId.
+ */
+public final class StandingOrderTokenBuilder {
```
yuhanniu: do we need a `fromTokenRequest` method?

PR #326, opened 2019-08-16
CAF

zixunjian:


core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -370,6 +412,19 @@ public TransferBuilder setProviderTransferMetadata(ProviderTransferMetadata meta
...
             return this;
         }
+
+        /**
+         * Optional. Sets whether CAF should be attempted before transfer.
+         *
+         * @param confirmFunds whether to attempt CAF before transfer
+         * @return builder
+         */
+        public TransferBuilder setConfirmFunds(boolean confirmFunds) {
```
yuhanniu: Are there relevant changes for TokenPayload?

zixunjian:
TransferPayload is not changed

tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -443,7 +443,11 @@ Member forAccessTokenInternal(String tokenId, boolean customerInitiated) {
                 .setTokenId(token.getId())
                 .setDescription(token
                         .getPayload()
-                        .getDescription());
+                        .getDescription())
+                .setConfirmFunds(token
```
yuhanniu: discussed offline, confirm_funds shouldn't be in the `TransferPayload`, revert. Same for other places.

PR #331, opened 2019-08-29
RD-3115

yuhanniu:
Setters for ultimateDebtor, ultimateCreditor and purposeCode

### DIFF
--- a/core/src/main/java/io/token/tokenrequest/TokenRequest.java
+++ b/core/src/main/java/io/token/tokenrequest/TokenRequest.java
@@ -34,7 +34,6 @@ import io.token.proto.common.token.TokenProtos.TokenRequestPayload.TransferBody;
 import io.token.proto.common.transferinstructions.TransferInstructionsProtos.ChargeBearer;
 import io.token.proto.common.transferinstructions.TransferInstructionsProtos.TransferDestination;
 import io.token.proto.common.transferinstructions.TransferInstructionsProtos.TransferEndpoint;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -382,6 +381,48 @@ public TransferBuilder setChargeBearer(ChargeBearer chargeBearer) {
                     .setChargeBearer(chargeBearer);
             return this;
         }
+
+        /**
+         * Optional. Sets the ultimate party to which an amount of money is due.
```
gianluca-token: "to which the money is due"?

zixunjian:
what does "ultimate debtor/creditor" mean?

core/src/main/java/io/token/tokenrequest/TokenRequest.java
yuhanniu: this is the ISO 20022 definition, does it look weird?

yuhanniu:
@zixunjian my understanding is for example if paypal is charging on behalf of a merchant, paypal is the creditor, the merchant is the ultimate creditor.

core/src/main/java/io/token/tokenrequest/TokenRequest.java
yuhanniu: changed, i think what you have makes more sense in this context.

PR #335, opened 2019-09-06
RD-3092: change return type for verifyEidas

nikanka-token:


tpp/src/main/java/io/token/tpp/rpc/Client.java
```diff
@@ -541,14 +542,14 @@ public NotifyStatus apply(TriggerStepUpNotificationResponse response) {
     /**
      * Verifies eIDAS certificate.
      *
-     * @param payload payload containing member id and the certificate
+     * @param payload payload containing member id, eIDAS alias and the certificate
      * @param signature payload signed with the private key corresponding to the certificate
      * @return a completable
```
Prateek-Gupta1: return Observable of VerifyEdiasResponse

PR #342, opened 2019-09-10
RD-3086: Add verifyEidas sample and test

nikanka-token:


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -1130,7 +1130,7 @@ public Account createTestBankAccountBlocking(double balance, String currency) {
      *     to the member</li>
      *     <li>The realmId of the alias is equal to the member's realmId</li>
      *</ul>
-     * @param payload payload containing the member id and the certificate in PEM format
+     * @param payload payload containing the member id and the base64 encoded eIDAS certificate
```
yuhanniu: just to be sure, when you say base64 encoded you mean the PEM format without the header and footer?

nikanka-token: yes, and it's exactly how Konsentus describes the input certificate format: https://docs.sandbox.konsentus-dev.com/api-reference/endpoints/psp-eidas.get.html

PR #347, opened 2019-09-11
RD-3123: add bulk transfers

gianluca-token:
java 8 refactoring


tpp/src/main/java/io/token/tpp/rpc/Client.java
```diff
@@ -442,17 +407,10 @@ public Transfer apply(GetTransferResponse response) {
      *
      * @param securityMetadata security metadata
      */
-    public void setSecurityMetadata(SecurityMetadata securityMetadata) {
+    private void setSecurityMetadata(SecurityMetadata securityMetadata) {
```
yuhanniu: why is this private and clearSecurityMetadata removed?

gianluca-token: It was unused outside of the class and clearSecurityMetadata was never used. Since this is Client.java I assume people won't call the methods directly?

PR #348, opened 2019-09-13
RD-3086: style fix, bump version

nikanka-token:


tpp-sample/src/main/java/io/token/sample/VerifyEidasSample.java
```diff
@@ -18,6 +18,15 @@
 
 public class VerifyEidasSample {
 
+    /**
+     * Creates a TPP member and verifies it using eIDAS certificate.
+     * @param client token client
```
gianluca-token: nit: \n before the first @param

PR #349, opened 2019-09-13
remove some java 8 to support older android systems

gianluca-token:


user/src/main/java/io/token/user/TransferTokenBuilder.java
```diff
@@ -385,7 +385,22 @@ public TransferTokenBuilder setReceiptRequested(boolean receiptRequested) {
      */
```
sibinlu: we should put a description here saying LocalDate is not supported before Android API 26 so please use the fallback function.

user/src/main/java/io/token/user/TransferTokenBuilder.java
```diff
@@ -385,7 +385,22 @@ public TransferTokenBuilder setReceiptRequested(boolean receiptRequested) {
...
+
+    /**
+     * Sets the execution date of the transfer. Used for future-dated payments.
+     *
+     * @param executionDate execution date
+     * @return builder
+     * @deprecated for use on older android systems
+     */
+    @Deprecated
+    public TransferTokenBuilder setExecutionDate(String executionDate) {
```
sibinlu: I don't see a reason we need to put it deprecated. If that is the only available api for Android developers. And they are the only user we have for user sdk now.

sibinlu: And I'm thinking actually no one will use the localdate API for now. I feel it is better to remove it if no one uses and reduces confusions.  But I will leave it to you.

user/src/main/java/io/token/user/Member.java
```diff
@@ -493,6 +494,58 @@ public TransferTokenBuilder createTransferTokenBuilder(TokenPayload tokenPayload
...
+     *
+     * @param amount individual transfer amount
+     * @param currency currency code, e.g. "USD"
+     * @param frequency ISO 20022 code for the frequency of the standing order:
+     *                  DAIL, WEEK, TOWK, MNTH, TOMN, QUTR, SEMI, YEAR
+     * @param startDate start date of the standing order
+     * @param endDate end date of the standing order
+     * @return standing order token builder
+     */
+    public StandingOrderTokenBuilder createStandingOrderTokenBuilder(
```
sibinlu: comments on the LocalDate function that LocalDate is not supported under Android API level 26 and please use the other endpoint.

PR #351, opened 2019-09-17
support filtering with bank feature map

sibinlu:


core/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -407,17 +409,18 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
             @Nullable Integer page,
             @Nullable Integer perPage,
             @Nullable String sort,
-            @Nullable String provider) {
+            @Nullable String provider,
+            @Nullable Map<String, String> bankFeaturesMap) {
```
gianluca-token: Map<String, Boolean>?

core/src/main/java/io/token/rpc/UnauthenticatedClient.java
```diff
@@ -429,7 +434,14 @@ public MemberRecoveryOperation apply(CompleteRecoveryResponse response) {
             request.setSort(sort);
         }
         if (provider != null) {
-            request.setProvider(provider);
+            request.getFilterBuilder().setProvider(provider);
+        }
+        if (bankFeaturesMap != null) {
+            Map<String, String> map = new HashMap();
```
gianluca-token: nit: can probably do 440 - 444 in one line with streams?

sibinlu: Collections class is in Java8

PR #362, opened 2019-10-04
get transaction by date

sibinlu:


gianluca-token:
That's the only one TPPs see

tpp-sample/src/main/java/io/token/sample/RedeemAccessTokenSample.java
```diff
@@ -60,6 +60,10 @@ public static Money redeemBalanceAccessToken(Member grantee, String tokenId) {
         PagedList<Transaction, String> transactions = accounts.get(0)
                 .getTransactionsBlocking(null, 10, STANDARD);
 
+        // Get the first 10 transactions by date
```
gianluca-token: "Get the 10 most recent transactions in the specified range"?

PR #375, opened 2019-11-01
Add setSource to TransferTokenBuilder

yuhanniu:


core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -397,6 +397,20 @@ public TokenRequest build() {
...
         }
 
+        /**
+         * Optional. Sets the source account to bypass account selection. May be required for
+         * some banks.
+         *
+         * @param source source
+         * @return builder
+         */
+        public TransferBuilder setSource(TransferEndpoint source) {
```
sibinlu: Is it for TransferBuilder only or we should add it to StandingOrderBuilder as well?

sibinlu: discussed offline. standing order builder already has this function.

PR #380, opened 2019-11-14
RD-3761: add getTransactions by date to representable

gianluca-token:


yuhanniu:
should they be implemented?

gianluca-token:
They were implemented in Member, they just needed to be added to the Representable interface.

PR #402, opened 2020-02-21
RD-3881: add getEidasVerificationStatus call

nikanka-token:


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -1211,6 +1212,17 @@ public Account createTestBankAccountBlocking(double balance, String currency) {
         return client.verifyEidas(payload, signature);
     }
 
+    /**
+     * Retrieves an eIDAS verification status by verificationId.
+     *
+     * @param verificationId verification id
+     * @return a status of the verification operation together with the certificate and alias value
+     */
+    public Observable<Gateway.GetEidasVerificationStatusResponse> getEidasVerificationStatus(
```
yuhanniu: nit: import

PR #409, opened 2020-03-05
RD-3764: implement createSigner(keyId) in android module

nikanka-token:
Do I need to bump the version? How?


android/src/main/java/io/token/security/AKSCryptoEngine.java
```diff
@@ -241,6 +256,32 @@ private String lookUpUnexpiredAliasInKeyStore(Level keyLevel) {
...
 
+    /**
+     * Looks up an unexpired alias in the keystore, by the keyId.
+     *
+     * @param keyId keyId
+     * @return alias
+     */
+    private String lookUpUnexpiredAliasInKeyStore(String keyId) {
+        try {
+            Enumeration<String> aliases = keyStore.aliases();
```
yuhanniu: the sdk is on java8 right, you can probably use `Stream`.

PR #424, opened 2020-05-06
PLT-109: add registerWithEidas call

nikanka-token:
To make it work, I also had to 
1) add possibility to create keys for different crypto types in TokenCryptoEngine;
2) use a key of a requested level or higher (if no key can be found for the requested level) in ClientAuthenticator;

I also formalized exceptions thrown by KeyStore when a key cannot be found or expired, to better catch exception it ClientAuthenticator when looking for an existing key.

### DIFF
--- a/core/src/main/java/io/token/rpc/ClientAuthenticator.java
+++ b/core/src/main/java/io/token/rpc/ClientAuthenticator.java
@@ -31,8 +31,10 @@ import com.google.common.base.Strings;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Message;
 import io.grpc.Metadata;
+import io.token.exceptions.KeyNotFoundException;
 import io.token.proto.common.security.SecurityProtos.CustomerTrackingMetadata;
 import io.token.proto.common.security.SecurityProtos.Key;
+import io.token.proto.common.security.SecurityProtos.Key.Level;
 import io.token.proto.gateway.Auth.GrpcAuthPayload;
 import io.token.rpc.interceptor.SimpleInterceptor;
 import io.token.security.CryptoEngine;
@@ -64,7 +66,7 @@ final class ClientAuthenticator<ReqT, ResT> extends SimpleInterceptor<ReqT, ResT
                 .setCreatedAtMs(now)
                 .build();
         Key.Level keyLevel = authenticationContext.getKeyLevel();
-        Signer signer = crypto.createSigner(keyLevel);
+        Signer signer = createSigner(keyLevel);
         String signature = signer.sign(payload);
 
         metadata.put(Metadata.Key.of("token-realm", ASCII_STRING_MARSHALLER), "Token");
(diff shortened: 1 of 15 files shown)


core/src/main/java/io/token/rpc/ClientAuthenticator.java
```diff
@@ -64,7 +66,7 @@ public void onStart(ReqT reqT, Metadata metadata) {
                 .setCreatedAtMs(now)
                 .build();
         Key.Level keyLevel = authenticationContext.getKeyLevel();
-        Signer signer = crypto.createSigner(keyLevel);
+        Signer signer = createSigner(keyLevel);
         String signature = signer.sign(payload);
 
         metadata.put(Metadata.Key.of("token-realm", ASCII_STRING_MARSHALLER), "Token");
```
ab666, line 72: Introduce a constant for the "token-realm"

core/src/main/java/io/token/rpc/ClientAuthenticator.java
```diff
@@ -104,4 +106,17 @@ public void onStart(ReqT reqT, Metadata metadata) {
     public void onHalfClose(ReqT req, Metadata headers) {
         // Ignore
     }
+
+    private Signer createSigner(Level minKeyLevel) {
+        Level keyLevel = minKeyLevel;
+        while (keyLevel.getNumber() > 0) {
```
ab666, line 112: move logic to CryptoEngine?

core/src/main/java/io/token/rpc/ClientAuthenticator.java
```diff
@@ -104,4 +106,17 @@ public void onStart(ReqT reqT, Metadata metadata) {
...
+            try {
+                return crypto.createSigner(keyLevel);
+            } catch (KeyNotFoundException e) {
+                // try a key for the next level
+                keyLevel = Level.forNumber(keyLevel.getNumber() - 1);
+            }
+        }
+        throw new KeyNotFoundException("Key not found for level " + minKeyLevel + " or higher.");
```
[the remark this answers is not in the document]
nikanka-token: Not for this one but I can add it - it just will be used only once - here.

core/src/main/java/io/token/security/UnsecuredFileSystemKeyStore.java
```diff
@@ -156,7 +160,7 @@ private SecretKey keyFromFile(File keyFile) {
         try {
             return codec.decode(keyFile.getName(), Files.toString(keyFile, Charsets.UTF_8));
         } catch (FileNotFoundException e) {
-            throw new IllegalArgumentException("Key not found: " + keyFile);
+            throw new KeyNotFoundException("Key not found: " + keyFile);
```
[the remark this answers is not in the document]
nikanka-token: Not for this one but I can add it - it just will be used only once - here.

core/src/main/java/io/token/rpc/ClientAuthenticator.java
nikanka-token: Not related to this PR and there are a lot of them there

PR #428, opened 2020-05-12
add registerWithEidasBlocking

nikanka-token:


tpp/src/main/java/io/token/tpp/TokenClient.java
```diff
@@ -450,6 +469,85 @@ public Member completeRecoveryWithDefaultRuleBlocking(
...
+     * @param privateKey
+     * @param keyStore
+     * @return a registered member
+     * @throws EidasRegistrationException
+     */
+    public Member registerWithEidasBlocking(
+            String bankId,
+            X509Certificate certificate,
+            PrivateKey privateKey,
+            KeyStore keyStore) {
```
nikanka-token, line 487: Still need key store here - to add the key. To avoid it, we need (for example) to make CryptoEngine to be able to not just generate keys, but also accept existing keys.

tpp/src/main/java/io/token/tpp/TokenClient.java
```diff
@@ -450,6 +469,85 @@ public Member completeRecoveryWithDefaultRuleBlocking(
...
+                        resp.getKeyId(),
+                        PRIVILEGED,
+                        new KeyPair(certificate.getPublicKey(), privateKey)));
+
+        // get the member
+        Member member = getMemberBlocking(memberId);
+        // wait until it is onboarded
+        String statusDetails = "";
+        EidasVerificationStatus verificationStatus = EIDAS_STATUS_PENDING;
+        long t = 1000;
```
nikanka-token, line 522: Have no idea how to handle this better. This is a naive version

nikanka-token: @ab666 reminding you to look at this and to think how to make it better

nikanka-token:
@ab666 I ended up not deleting CryptoEngine.createSignerForLevelAtLeast, as it is published and somebody can already be using it. Also, I think it is a nice thing to have tbh.

Also I do not like the fact that we have two similar methods: registerWithEidas and createMemberWithEidas which do the same (only the second one is blocking), but the names are very different. I would even say that their names should be swapped to better match what they do. Do you think this is fine?

tpp/src/main/java/io/token/tpp/security/EidasCryptoEngineFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public EidasCryptoEngineFactory(EidasKeyStore keyStore) {
+        this.keyStore = keyStore;
+    }
+
+    @Override
+    public CryptoEngine create(String memberId) {
+        return new TokenCryptoEngine(memberId, keyStore, CryptoType.RS256);
+    }
+
```
ab666, line 41: delete

ab666: I mean delete the extra line

tpp/src/main/java/io/token/tpp/TokenClient.java
ab666: Take a look at this library https://github.com/nurkiewicz/async-retry if it is of any help

nikanka-token: @ab666 I decided no to use this library, as:
1) this is a single place where we need it and it's just several lines of code
2) last release was in 2015 and it is for java 7/8.
3) it is only for java - not suited for other sdks.
4) I do not see a way to determine the time after which to give up  - only the max number of attempts.
5) it retries on exceptions, so I would have to wrap the status into exception, which might result in more code than we have now.

I created a util function to do this, because a separate function is easier to test: https://github.com/tokenio/sdk-java/pull/432/commits/bae01c1108dbadbfe8113567c6ef76c5fac5acbf

tpp/src/main/java/io/token/tpp/security/EidasCryptoEngineFactory.java
nikanka-token: I got this :) I asked you to read [another comment](https://github.com/tokenio/sdk-java/pull/428#issuecomment-629479796)

PR #432, opened 2020-05-21
PLT-185: add createMemberWithEidas

nikanka-token:
Same as https://github.com/tokenio/sdk-java/pull/428 but with a util function to make a status check call until the status is not pending.


tpp/src/main/java/io/token/tpp/util/Util.java
```diff
@@ -145,4 +147,91 @@ public static void verifySignature(
...
+            throw new IllegalArgumentException(
+                    "All time arguments and the backOffFactor should be non-negative.");
+        }
+        long totalTime = 0;
+        T result = function.call();
+        while (retryIf.test(result)) {
+            if (totalTime >= timeoutMs) {
+                return result;
+            }
+            Thread.sleep(waitTimeMs);
```
ab666, line 188: Can you use a scheduled executor instead?

PR #434, opened 2020-05-26
PLT-271: require redirect url, ref id on token request

gianluca-token:


core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -374,6 +374,12 @@ public T setTokenExpiration(long tokenExpiration) {
          * @return TokenRequest instance
          */
         public TokenRequest build() {
+            if (this.requestPayload.getRedirectUrl().isEmpty()) {
+                throw new IllegalArgumentException("Redirect URL required");
+            }
+            if (this.requestPayload.getRefId().isEmpty()) {
+                this.requestPayload.setRefId(Util.generateNonce());
```
yuhanniu, line 381: Maybe we should stop dong this so that the sdk and the rest api behave more consistently.

PR #480, opened 2020-10-02
plt-1010: bump sdk version to include the new status

zzltoken:


sibinlu (approved):
build it before merging to avoid dependencies conflict.

PR #493, opened 2020-11-18
PLT-1326: add InitiateBankAuthorization

gianluca-token:


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -1247,6 +1247,54 @@ public Account createTestBankAccountBlocking(double balance, String currency) {
...
     }
 
+    /**
+     * Initiate authorization process with the source bank, for an existing token request.
+     *
+     * @param tokenRequestId token request ID
+     * @param trackingMetadata customer tracking metadata
+     * @return initiation result
+     */
+    public Observable<InitiateBankAuthorizationResult> initiateBankAuthorization(
```
yuhanniu, line 1257: I have the same dilemma here. I think maybe we should just use the rpc response here so that the client don't lose the one-of attribute and can use switch on the response.

gianluca-token: Sure, didn't realize that pattern was already established.

yuhanniu (approved):

PR #495, opened 2020-11-24
PLT-1346: ability to publish uber jars

ab666:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Sep 10 15:50:33 PDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
```
javidang, line 3: will it be fine to use 6.7 here ? I had troubles setting up new projects 6+ versions.

javidang (approved):

build.gradle
```diff
@@ -32,5 +29,5 @@ allprojects {
     }
 
     group = 'io.token.sdk'
-    version = '2.12.24'
+    version = '2.12.25-SNAPSHOT'
```
drobinson123, line 32: -SNAPSHOT?

ab666: oops

gradle/wrapper/gradle-wrapper.properties
ab666: should be fine

PR #496, opened 2020-11-24
SRE-247: Allow custom environments

drobinson123:


core/src/main/java/io/token/TokenClient.java
```diff
@@ -690,7 +691,10 @@ private static String getTraceId() {
         }
 
         public String url() {
-            return url;
+            if (url == "TOKEN_URL")
```
ab666, line 694: `"TOKEN_URL".equals(url)`

PR #514, opened 2021-02-16
updates to token APIs, remove deprecated methods

gianluca-token:
Important changes
- including transferId in TokenRequestResult
- adding redeemToken(String tokenId)
- deprecating TokenRequestCallback, parseTokenCallbackUrl (they should just get the tokenRequestId query param)

The rest is just deleting stuff that was already marked as deprecated, converting Functions to lambdas, and updating 2020 -> 2021.


javidang:
Hi @gianluca-token , I cannot see the changes for `TokenRequestCallback` or `parseTokenCallbackUrl`. should I search with different keyword ?

gianluca-token:
Sure, it's deprecating methods like this https://github.com/tokenio/sdk-java/pull/514/files#diff-ee19754cdd04f202a1ea8648881ce36f2b1fd0da46433a801f2eb19f468fd7ddL654 (probably needed to expand TokenClient.java before searching) and adding the comment: "Deprecated: The callback will contain request-id only. tokenId and state will be removed."

javidang:
I see, yeah that was the only file that haven't loaded. overall LGTM.

javidang (approved):

PR #556, opened 2021-08-19
PLT-2155 adding transaction history days

ramesh-token:


user/src/main/java/io/token/user/AccessTokenBuilder.java
```diff
@@ -117,6 +117,8 @@ public static AccessTokenBuilder fromTokenRequest(TokenRequest tokenRequest) {
         if (tokenRequest.getRequestPayload().hasActingAs()) {
             builder.setActingAs(tokenRequest.getRequestPayload().getActingAs());
         }
+        builder.getAccessBuilder().setTransactionHistoryDays(tokenRequest.getRequestPayload().getAccessBody().getTransactionHistoryDays());
```
sibinlu, line 120: We need to provide a set endpoint for users to set TransactionHistoryDays with the builder. e.g. setRefId.

core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -443,6 +443,18 @@ public TokenRequest build() {
                             .setAccountResourceList(list)
                             .build());
         }
+
+        /**
+         * optional set the transaction history days.
```
sibinlu, line 448: Please follow the other Javadoc style.

core/src/main/java/io/token/tokenrequest/TokenRequest.java
```diff
@@ -443,6 +443,18 @@ public TokenRequest build() {
...
+         * @param transactionHistoryDays transactionHistoryDays
```
sibinlu, line 449: here as well.

core/src/main/java/io/token/tokenrequest/TokenRequest.java
ramesh-token: Updated java doc.

core/src/main/java/io/token/tokenrequest/TokenRequest.java
ramesh-token: updated java doc.

sibinlu (approved):

sibinlu (approved):
Reminder: merge it only if your changes are in production to avoid any use from our clients.

PR #560, opened 2022-01-31
PLT-2585: added the createSubTpp method

rohit-Jadon:


tpp/src/main/java/io/token/tpp/rpc/Client.java
```diff
@@ -276,6 +278,31 @@ private void useAccessToken(
...
+                .map(response -> response.getSubTppId());
+    }
+
+    /**
+     * Deletes a SubTpp.
+     *
+     * @param deleteSubTppRequest delete subTpp request
+     */
+    public void deleteSubTpp(DeleteSubTppRequest deleteSubTppRequest) {
+        toObservable(gateway
```
sibinlu, line 300: check the example https://github.com/tokenio/sdk-java/blob/b79c1b187536172e1bb66212f19c8a834ae94eed/tpp/src/main/java/io/token/tpp/rpc/Client.java#L158

sibinlu (approved):

PR #561, opened 2022-02-08
PLT-2579: added required methods for subTpp

rohit-Jadon:


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -702,6 +706,68 @@ public void deleteSubTppBlocking(DeleteSubTppRequest deleteSubTppRequest) {
...
+     *
+     * @param retrieveSubTppsRequest retrieve subTpp request
+     * @return List of subTpp
+     */
+    public List<SubTpp> retrieveSubTppsBlocking(RetrieveSubTppsRequest retrieveSubTppsRequest) {
+        return retrieveSubTpps(retrieveSubTppsRequest).blockingSingle();
+    }
+
+    /**
+     * Retrieves SubTpps.
```
sibinlu, line 740: suggestion: Retrieves root SubTpps.
this is easier to distinguish from the other method.

sibinlu (approved):

PR #568, opened 2022-09-01
PLT-2840: added the required methods for fraud-control service

rohit-Jadon:


tpp/src/main/java/io/token/tpp/Member.java
```diff
@@ -49,14 +49,23 @@
 import io.token.proto.common.transfer.TransferProtos.Transfer;
 import io.token.proto.common.transferinstructions.TransferInstructionsProtos.TransferDestination;
 import io.token.proto.common.webhook.WebhookProtos.Webhook.Config;
+import io.token.proto.gateway.Gateway;
```
yusuf-aps, line 52: please remove if not used.

yusuf-aps (approved):

PR #577, opened 2023-12-20
L2S-637: Upgrading version of token rpc.

emircankilinc:
Upgrading version of token rpc.


emircankilinc:
We need to fix the issue on netty lib-rpc asap: https://tokenio.atlassian.net/browse/PLT-4170

rohit-Jadon (approved):

allenkunju (approved):

PR #579, opened 2024-01-04
PLT-4170: Resolution of Dependency Selection Issue

emircankilinc:
Resolution of Dependency Selection Issue for tpp and user sample.


build.gradle
```diff
@@ -2,7 +2,6 @@ buildscript {
     repositories {
         mavenLocal()
         google()
-        jcenter()
```
emircankilinc, line 5: Builds will no longer be able to resolve artifacts from JCenter after February 1st, 2022

tpp-sample/build.gradle
```diff
@@ -1,11 +1,29 @@
...
+} else if (OperatingSystem.current().isWindows()) {
+    platformSpecificClassifier = 'windows-x86_64'
+}
+
 dependencies {
     compile project(':tpp')
-    compile 'junit:junit:4.12'
+    compile 'junit:junit:4.13.1'
     compile group: 'io.token.rpc', name: 'tokenio-rpc-client-lite-netty', version: ver.tokenRpc
+    testImplementation group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: ver.nettyBoringssl, classifier: platformSpecificClassifier
```
emircankilinc, line 26: This is not transitive so needed to add for fixing building tpp-sample and user-sample.

tpp-sample/build.gradle
```diff
@@ -1,11 +1,29 @@
...
+    if (osArch == "x86_64") {
+        platformSpecificClassifier = 'osx-x86_64' // Required for Mac with intel chip
+    } else {
+        platformSpecificClassifier = 'osx-aarch_64' // Required by Apple Silicon MacBooks
+    }
+} else if (OperatingSystem.current().isLinux() && osArch == "aarch64") {
+    platformSpecificClassifier = 'linux-aarch_64' // Required by ARM64 Linux
+} else if (OperatingSystem.current().isWindows()) {
+    platformSpecificClassifier = 'windows-x86_64'
+}
```
emircankilinc, line 20: I'm open to moving this to the parent build.gradle file. Before making a suggestion, could you please try this locally? I attempted to do the same for both tpp-sample and user-sample, but it didn't work on my end. Thanks.

tpp-sample/build.gradle
rohit-Jadon: but my question here is why the test cases where working fine before? I remember when i updated the lib-rpc version to 2.4.2 it was working fine

tpp-sample/build.gradle
rohit-Jadon: try moving this to core package in sdk-jav build.gradle and see if it works. Cause these tpp-sample is inheriting from core

tpp-sample/build.gradle
emircankilinc: We may want to check for changes or dependency upgrades between version 2.4.2 and the latest release. Some of them may not be compatible so it affects the deps. Currently it is hard to say what is wrong or why it does not work without comparing the changes.

tpp-sample/build.gradle
```diff
@@ -1,11 +1,29 @@
+import org.gradle.internal.os.OperatingSystem
+
 apply plugin: 'io.token.gradle.publish'
 apply plugin: 'io.token.gradle.lib'
 apply plugin: 'io.token.gradle.project'
 
+def platformSpecificClassifier = 'linux-x86_64'
+def osArch = System.getProperty("os.arch")
+
+if (OperatingSystem.current().isMacOsX()) {
```
allenkunju, line 10:
```suggestion
dev currentOs = OperatingSystem.current()
if (currentOs.isMacOsX()) {
```
nit: refactor

emircankilinc: make sense for better readability

allenkunju (approved):

